PR #133, opened 2024-11-17
Reorder dataservice publish form

geoffreyaldebert:
- [x] update: translation for dataservice
- [x] reorder fields
- [x] finish checkup for dataservice


nicolaskempf57:
I think that the reorder won't work because of the way the datasets are fetched.
@ThibaudDauce knows more about this

geoffreyaldebert:
that is not the reorder of the datasets sorry but the reorder of fields in the form. this one os testednand work for me. reordering datasets is another subject not in this PR

ThibaudDauce:
@geoffreyaldebert  Just a quick look and I don't understand why there is a new `AdminDataservicesTable` file? :-) But I will pull it and test it locally to try to understand where it is used…

components/AdminDataservicesTable/AdminDataservicesTable.vue
ThibaudDauce, line 1: I'm not sure to understand where this is used…

ThibaudDauce (approved):

PR #134, opened 2024-11-17
Feat/remove objects in edit mode

geoffreyaldebert:
- add button "dangerzone" for datasets, reuses and dataservices in edit mode


components/Datasets/AdminUpdateDatasetPage.vue
```diff
@@ -73,4 +139,17 @@ const save = async () => {
...
   }
 }
+
+async function deleteDataset() {
+  loading.value = true
+  try {
+    await $api(`/api/1/datasets/${route.params.id}`, {
+      method: 'DELETE',
+    })
+    navigateTo(`${config.public.apiBase}/en/beta/admin/organizations/${route.params.oid}/datasets`, { external: true })
```
nicolaskempf57, line 149: The correct redirect logic should be something like this : https://github.com/datagouv/front-end/blob/feat/remove-objects-edit-mode/pages/beta/admin.vue#L71

apiBase is for api or links not already moved to this new front-end

components/Datasets/AdminUpdateDatasetPage.vue
```diff
@@ -7,6 +7,56 @@
...
+            color="red"
+            size="sm"
+            type="secondary"
+            :aria-controls="modalId"
+            :icon="RiDeleteBin6Line"
+            @click="openDeleteModal"
+          >
+            {{ $t('Delete') }}
+          </BrandedButton>
+          <Modal
```
nicolaskempf57, line 32: There is a `ModalWithButton` component that is dealing with the id and open / close function to avoid the duplication

components/Dataservices/AdminUpdateDataservicePage.vue
```diff
@@ -71,4 +137,30 @@ const save = async () => {
...
+  try {
+    await $api(`/api/1/datasets/${route.params.id}`, {
+      method: 'DELETE',
+    })
+    navigateTo(`${config.public.apiBase}/en/beta/admin/organizations/${route.params.oid}/datasets`, { external: true })
+  }
+  finally {
+    loading.value = false
+  }
+}
```
ThibaudDauce, line 165: Code not used
```suggestion
```

components/Dataservices/DescribeDataservice.vue
```diff
@@ -425,7 +425,6 @@
           <slot />
         </div>
       </div>
-      <div class="h-64" />
```
ThibaudDauce, line 428: This line is useful to have a little bit of margin bottom below the forms (it could be done with a margin-bottom on the container I think, it could be better but maybe keep it and change this in another PR?)

components/Datasets/DescribeDataset.vue
```diff
@@ -462,7 +462,6 @@
           </button>
         </div>
       </div>
-      <div class="h-64" />
```
ThibaudDauce, line 465: Same here :-)

ThibaudDauce (approved):

components/Dataservices/DescribeDataservice.vue
ThibaudDauce: Let's merge then :-) I will add it back if necessary

components/Datasets/AdminUpdateDatasetPage.vue
geoffreyaldebert: I created an issue to do that for all modals (i just copied an existing one for account deletion)

components/Datasets/AdminUpdateDatasetPage.vue
geoffreyaldebert: thanks indeed it is better

geoffreyaldebert:
When you remove a dataservice, you go back to admin dataservice table and your dataservice is not showing that it is deleted (but it is). Is is due to datagouv/components that does not take `deleted` metadata into property and we can't display it in this repo. I create an issue but let's merge that.

PR #144, opened 2024-11-18
Use localPath on administration header button

maudetes:


components/SiteHeader/SiteHeader.vue
```diff
@@ -72,7 +72,7 @@
                 </li>
                 <li>
                   <a
-                    href="/en/beta/admin/"
+                    :href="localePath('/beta/admin/')"
```
nicolaskempf57, line 75: I think we should use `<NuxtLinkLocale>` instead of a `<a>` with a localePath

PR #158, opened 2024-11-18
feat: add checkbox draft in edit mode

geoffreyaldebert:


ThibaudDauce:
I'm not a big fan of the checkbox here (but more for design purposes than code…). I think it's better to have a banner like the delete one with "Switch to draft" or "Undraft - Publish this dataset". But not sure what @agarrone think of this (I know he designed the checkbox in Figma ^^)

ThibaudDauce:
And even with the checkbox, I think it's clearer with a radio button with explicit choices "Draft" and "Public" than a checkbox we don't always understand the way it's suppose to be checked or not.

geoffreyaldebert:
That was on the UI mockup but I agree with you. Maybe we can improve that later ?

ThibaudDauce (approved):

PR #196, opened 2024-12-05
SiteLogo as text

ThibaudDauce:


components/BrandedButton/BrandedButton.vue
```diff
@@ -1,11 +1,11 @@
 <template>
   <component
-    :is="as"
+    :is="href ? 'a': 'button'"
```
nicolaskempf57, line 3: I think we should use a `NuxtLinkLocale` here instead of a `a` to avoid a full page load

ThibaudDauce: I don't know how to do it… https://github.com/nuxt-modules/i18n/issues/3257

ThibaudDauce: I didn't find the correct import for `NuxtLinkLocale`

nicolaskempf57: The correct import is `import { NuxtLinkLocale } from '#components'` but it is indeed quite hard to find and not autocompleted

assets/fonts/gothic.woff
nicolaskempf57, line 1: Do we have the licences for these somewhere and the right to use it ? The only open source version of them I found are from [Ghostscript released under AGPL](https://git.ghostscript.com/?p=ghostpdl.git;a=blob;f=LICENSE;h=572496a331cefa0933a9e997374f5e7dde3b95c8;hb=0e50ac9dfcff795ec69cb646eea151bd57f0dd1c#l13) so it seems not compatible with our MIT licence.

ThibaudDauce: For the legal team it's the OFL license that apply

nicolaskempf57 (approved):

PR #197, opened 2024-12-06
Add read only mode

nicolaskempf57:


layouts/default.vue
```diff
@@ -1,6 +1,10 @@
 <template>
-  <div>
+  <div
+    class="group/layout"
+    :data-with-container="true"
```
maudetes, line 4: What is this used for?

maudetes (approved):
Thank you! Looks good :pray:

layouts/default.vue
nicolaskempf57: The new banner component must should show its content at the same position that our layout : at the edge of the screen if we have a fluid layout or at the current container position if the page is using a container.
We can add the information as data attributes and [target it with CSS](https://github.com/datagouv/front-end/pull/197/files#diff-1e0715102e48517e4d89049836cd321633ffffd1e8f2d9957f29cb86a4733b7fR12) to provide the correct class only when required.

nuxt.config.ts
```diff
@@ -67,6 +67,8 @@ export default defineNuxtConfig({
       changeEmailPage: 'change-email',
       changePasswordPage: 'change',
 
+      readOnlyMode: false,
```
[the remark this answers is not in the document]
nicolaskempf57: no but at least our deployment pipeline can make one config from the other

PR #198, opened 2024-12-10
Improve profile page

nicolaskempf57:
Closes https://github.com/datagouv/front-end/issues/177
Parts of https://github.com/datagouv/front-end/issues/38

### DIFF
--- a/components/User/AdminUserProfilePage.vue
+++ b/components/User/AdminUserProfilePage.vue
@@ -278,6 +278,12 @@ const passwordId = useId()
 
 const loading = ref(false)
 
+watchEffect(() => {
+  if (me.value.about === null) {
+    me.value.about = ''
+  }
+})
+
 async function updateMe() {
   loading.value = true
   try {
--- a/components/User/AdminUserProfilePage.vue
+++ b/components/User/AdminUserProfilePage.vue
@@ -80,6 +80,25 @@
         :label="$t('Website')"
         type="url"
       />
+      <UploadGroup
+        :label="$t('Profile picture')"
+        type="drop"
+        accept=".jpeg, .jpg, .png"
+        :hint-text="$t('Max size: 4Mo. Accepted formats: JPG, JPEG, PNG')"
+        :show-label="true"
+        @change="setFiles"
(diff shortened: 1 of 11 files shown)


components/User/AdminUserProfilePage.vue
```diff
@@ -80,6 +80,25 @@
         :label="$t('Website')"
         type="url"
       />
+      <UploadGroup
+        :label="$t('Profile picture')"
+        type="drop"
+        accept=".jpeg, .jpg, .png"
+        :hint-text="$t('Max size: 4Mo. Accepted formats: JPG, JPEG, PNG')"
```
maudetes, line 87: Is the max size actually applied?

components/User/AdminUserProfilePage.vue
```diff
@@ -278,8 +298,34 @@ const passwordId = useId()
...
+  if (!profilePicture.value) return null
+  if (typeof profilePicture.value === 'string') return profilePicture.value
+  return URL.createObjectURL(profilePicture.value)
+})
+
 async function updateMe() {
   loading.value = true
+  if (profilePicture.value) {
+    try {
+      await uploadProfilePicture(profilePicture.value)
```
maudetes, line 323: Do we want to update the avatar at the top of the page upon upload?

maudetes (approved):
:rocket:

components/User/AdminUserProfilePage.vue
nicolaskempf57: It should update both the avatar on the admin page and small one inside the header

components/User/AdminUserProfilePage.vue
nicolaskempf57: This is not a native thing we can do like the formats, so it's never applied on the front-end.

But we could / should work on a better image upload experience with cropping like https://pqina.nl/filepond/

components/User/AdminUserProfilePage.vue
maudetes: Indeed, it is a case after a reinstall!

PR #199, opened 2024-12-10
Fix license combobox

nicolaskempf57:


maudetes (commented):
I'm confused about what we're trying to fix :p

nicolaskempf57:
> I'm confused about what we're trying to fix :p

The current implementation isn't working for license that are available for both the public services and the other organizations, the hover effect is missing

ThibaudDauce (approved):
Seems fine! The checkmark is much better :D 

image

Is it missing a margin between checkmark and label in frequencies select?

nicolaskempf57:
> Is it missing a margin between checkmark and label in frequencies select?

Indeed, now using a gap to work with this case too

PR #200, opened 2024-12-11
Fix missing full width for dataset selector

nicolaskempf57:


components/SearchableSelect.vue
```diff
@@ -218,7 +219,7 @@ watchEffect(async () => {
   if (savedQuery === query.value) {
     suggestedOptions.value = options
   }
-})
+}, { debounce: 400, maxWait: 800 })
```
maudetes, line 222: Isn't it slightly too much?

maudetes (approved):
:rocket:

components/SearchableSelect.vue
nicolaskempf57: I found that 300 was firing between 2 keystrokes so I changed it to 400 but I can reduce it back to 300

PR #203, opened 2024-12-11
Add loaders and show content when loading

nicolaskempf57:


ThibaudDauce:
I'm not sure it solves the problem of the jumping window? I was imagining more something like this https://github.com/datagouv/front-end/pull/209/ (I've only converted the org datasets page if you want to try and tell me what you think of this).

nicolaskempf57:
Done in https://github.com/datagouv/front-end/pull/209

PR #206, opened 2024-12-12
Fix danger zones misplaced

nicolaskempf57:


ThibaudDauce (approved):
A few classes are off but otherwise good for me :-)

PR #214, opened 2024-12-17
Show deleted dataservices status

ThibaudDauce:
I think it fixes https://github.com/datagouv/front-end/issues/136 but I don't understand the problem in the issue… @geoffreyaldebert maybe?


nicolaskempf57 (approved):

nicolaskempf57:
I think the other part of the issue is that dataservices use `deleted_at` and not `deleted` like the other objects

PR #218, opened 2024-12-18
Enable search on reuses

ThibaudDauce:


components/Reuses/AdminReusesPage.vue
```diff
@@ -29,21 +29,18 @@
...
-          v-if="pageData && pageData.total"
-          class="subtitle subtitle--uppercase fr-m-0"
-        >
+        <h2 class="subtitle subtitle--uppercase fr-m-0">
           {{ t('{n} reuses', pageData.total) }}
         </h2>
       </div>
       <div class="fr-col-auto fr-grid-row fr-grid-row--middle space-x-6">
-        <!-- <AdminInput
+        <AdminInput
```
nicolaskempf57, line 39: I think I prefer the dataset way with the input still shown when the search yield no results

nicolaskempf57: Also, I think we can change the AdminInput to `type="search"`

ThibaudDauce: Yes, me too, but I think I need to leave the `{{ t('{n} reuses', pageData.total) }}` too because otherwise the input seems to be floating at the right alone…

Could we merge this and then I do another PR to restore the search on empty input for all the screens?

nicolaskempf57: sure

PR #223, opened 2024-12-19
Add posts

nicolaskempf57:
This PR doesn't contain the public footer that I'll be working on next


ThibaudDauce (commented):
I think there is some translations missing. And maybe add some bottom padding to the page to allow scrolling past the article?

The HTML body type doesn't work, I don't know if it's wanted?

nicolaskempf57:
> I think there is some translations missing. And maybe add some bottom padding to the page to allow scrolling past the article?

Indeed the translation was almost the same as the admin one so I remove one during the conflict resolution. Now, it's really the same one as the admin page. 

> The HTML body type doesn't work, I don't know if it's wanted?

Oh you're right, I didn't had these in my local setup. Now it's better.

ThibaudDauce (approved):
Need to check on dev with the correct images but seems fine! (I don't have images on my local setup :-) )

pages/posts/[id].vue
```diff
@@ -0,0 +1,71 @@
...
+      :content="post.content"
+      :min-heading="2"
+    />
+    <div
+      v-else
+      :class="markdownClasses"
+      v-html="post.content"
+    />
+  </div>
+</template>
```
Comment (maudetes) [inline pages/posts/[id].vue:59]: Could we maybe add a `Edit` button for mega-admin users? :p

PR #224, opened 2024-12-19
Add post forms

ThibaudDauce:


pages/beta/admin/site/posts.vue
```diff
@@ -78,7 +78,14 @@
               </td>
               <td>{{ formatDate(post.created_at) }}</td>
               <td>{{ formatDate(post.last_modified) }}</td>
-              <td />
+              <td>
```
nicolaskempf57, line 81: Table header shouldn't be empty now that this column has content

pages/beta/admin/posts/[id]/index.vue
```diff
@@ -0,0 +1,54 @@
...
+import DescribePost from '~/components/Posts/DescribePost.vue'
+import type { Post, PostForm } from '~/types/posts'
+import { toApi } from '~/utils/posts'
+
+const { t } = useI18n()
+const { $api, $fileApi } = useNuxtApp()
+const { toast } = useToast()
+
+const route = useRoute()
+const url = computed(() => `/api/1/posts/${route.params.id}`)
```
Comment (nicolaskempf57) [inline pages/beta/admin/posts/[id]/index.vue:23]:
```suggestion
const url = computed(() => `/api/1/posts/${route.params.id}/`)
```

pages/beta/admin/posts/[id]/content.vue
```diff
@@ -0,0 +1,46 @@
...
+<script setup lang="ts">
+import PostContentForm from '~/components/Posts/PostContentForm.vue'
+import type { Post, PostForm } from '~/types/posts'
+
+const { t } = useI18n()
+const { $api } = useNuxtApp()
+const { toast } = useToast()
+
+const route = useRoute()
+const url = computed(() => `/api/1/posts/${route.params.id}`)
```
Comment (nicolaskempf57) [inline pages/beta/admin/posts/[id]/content.vue:22]:
```suggestion
const url = computed(() => `/api/1/posts/${route.params.id}/`)
```

components/Posts/DescribePost.vue
```diff
@@ -0,0 +1,138 @@
...
+      :class="{ 'fr-grid-row--right': type === 'update', 'justify-between': type === 'create' }"
+    >
+      <BrandedButton
+        v-if="type === 'create'"
+        color="secondary"
+        @click="$emit('previous')"
+      >
+        {{ $t('Previous') }}
+      </BrandedButton>
+      <BrandedButton
```
nicolaskempf57, line 91: This one is missing a loader

pages/beta/admin/site/posts.vue
```diff
@@ -78,7 +78,14 @@
...
+                <NuxtLinkLocale
```
nicolaskempf57, line 82: Do we want to link the post title to the edit page too to make it work like the other pages ?

nicolaskempf57 (approved):

nicolaskempf57 (approved):
Some small changes but overall this is great !

pages/beta/admin/site/posts.vue
ThibaudDauce: I think we need to do a pass on this on all tables. Do we want to always link to the update page or to the public page and add a edit button in the last column? Need to check this with @agarrone

components/Posts/DescribePost.vue
ThibaudDauce: I don't see a loader in `DescribeDataset` too… And this component is not the one doing the update so I may need to think of the best solution to do this… Could we postpone?

pages/beta/admin/site/posts.vue
ThibaudDauce: Done in 180265e

PR #232, opened 2024-12-31
Force markdown img lazy loading

nicolaskempf57:
Code is taken from https://github.com/potato4d/rehype-plugin-image-native-lazy-loading to avoid pulling a new dependency for a single function


ThibaudDauce (approved):
Maybe add a comment with a link to the original source code? (the tagged/commited link to keep the history?)

PR #234, opened 2024-12-31
Add job page with items list

ThibaudDauce:
This is messier than we want because we cannot have a simple `/harvest-jobs/[id]` page. The API doesn't provide a way to get back to the source from the job ID :-( So we need to be inside the source folder and inheritate the layout which is not exactly what we want…

### DIFF
--- a/pages/beta/admin/harvesters/[id].vue
+++ b/pages/beta/admin/harvesters/[id].vue
@@ -26,7 +26,25 @@
             {{ t('Harvesters') }}
           </NuxtLinkLocale>
         </li>
-        <li>
+        <template v-if="job">
+          <li>
+            <NuxtLinkLocale
+              class="fr-breadcrumb__link"
+              :to="getHarvesterAdminUrl(harvester)"
+            >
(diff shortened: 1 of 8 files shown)


pages/beta/admin/harvesters/[id].vue
```diff
@@ -119,6 +137,13 @@ const { toast } = useToast()
 const route = useRoute()
 const url = computed(() => `/api/1/harvest/source/${route.params.id}`)
 const { data: harvester } = await useAPI<HarvesterSource>(url, { lazy: true })
+const job = ref<HarvesterJob | null>(null)
+watchEffect(async () => {
+  if (!harvester.value) return
+  if (!route.params.jobid) return
+
+  job.value = await $api(`/api/1/harvest/job/${route.params.jobid}`)
```
Comment (nicolaskempf57) [inline pages/beta/admin/harvesters/[id].vue:145]:
```suggestion
  job.value = await $api(`/api/1/harvest/job/${route.params.jobid}/`)
```

nicolaskempf57 (approved):
Nice !

PR #240, opened 2025-01-08
Add new post page

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1626


ThibaudDauce (approved):
:+1: 

I think it misses some tranlations in the form?

PR #242, opened 2025-01-09
Add editButton to posts

nicolaskempf57:


pages/posts/[id].vue
```diff
@@ -1,33 +1,41 @@
...
+            class="fr-breadcrumb__link"
+            aria-current="page"
+          >
+            {{ post.name }}
+          </a>
+        </li>
+      </Breadcrumb>
+      <div v-if="isAdmin(me)">
+        <EditButton
+          :id="(route.params.id as string)"
```
Comment (ThibaudDauce) [inline pages/posts/[id].vue:34]: This link to the edit page with the post slug in URL (it's working because the API return the post anyway but the aria-current=page is not because it's comparing slug with ID…) Not a big deal for now but maybe add a ticket on it.

components/BrandedButton/EditButton.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <BrandedButton
+    color="danger"
```
ThibaudDauce, line 3: Danger seems a little bit overkill for this kind of action but why not :-)

ThibaudDauce (approved):
Only minor comments

components/BrandedButton/EditButton.vue
nicolaskempf57: This is the current color on udata-front, but I agree that we could change it later

PR #255, opened 2025-01-14
Add contact points to datasets

ThibaudDauce:


nicolaskempf57 (approved):
I found an unrelated issue so I added it here : https://github.com/datagouv/front-end/issues/257

PR #271, opened 2025-01-24
Add organizations

nicolaskempf57:
This adds the page as its own component to see what it could be in a future component package and / or layer

Closes https://github.com/datagouv/data.gouv.fr/issues/1636

This is the design, but we won't do the new title and headline until the new datasets/ and reuses/ pages

image

### DIFF
--- a/components/Organization/ListPage.vue
+++ b/components/Organization/ListPage.vue
@@ -60,7 +60,7 @@
     </div>
   </div>
   <LoadingBlock :status>
-    <div class="grid gap-2 grid-cols-3 mb-16">
+    <div class="grid gap-4 grid-cols-3 mb-16">
       <OrganizationCard
         v-for="organization in organizations.data"
         :key="organization.id"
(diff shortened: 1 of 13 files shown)


components/Organization/ListPage.vue
```diff
@@ -0,0 +1,122 @@
...
+        :for="selectId"
+        class="flex-none text-sm"
+      >{{ $t('Sort by:') }}</label>
+      <select
+        :id="selectId"
+        v-model="sort"
+        class="fr-select !shadow-input"
+      >
+        <option value="">
+          Relevance
```
maudetes, line 45: This should be a translation string :)

pages/organizations/index.vue
```diff
@@ -0,0 +1,65 @@
...
+    query: {
+      ...route.query,
+      q: q.value,
+      page: page.value,
+      sort: sort.value,
+    },
+  })
+}
+
+const { data: organizations, status } = await useAPI<PaginatedArray<Organization>>(`/api/1/organizations/`, { params:
```
maudetes, line 57: I think we should use the api/2 for the front page organization?

maudetes (approved):
Nice! Do you have a figma reference?

I don't have any opinion on the adding the organization in a component?

PR #273, opened 2025-01-29
Fix i18n date

nicolaskempf57:
Replace datagouv/components date functions with native ones


ThibaudDauce (changes requested):
On http://dev.local:3000/fr/beta/admin/site/users/ I have:

> RangeError: date value is not finite in DateTimeFormat.format()

nicolaskempf57:
> On http://dev.local:3000/fr/beta/admin/site/users/ I have:
> 
> > RangeError: date value is not finite in DateTimeFormat.format()

I added a `null` case to the function to avoid the error. And I fixed the users page using the wrong attribute

ThibaudDauce (approved):

PR #277, opened 2025-01-30
Add unselect button to searchable select

ThibaudDauce:


nicolaskempf57:
I find it weird that the use can't click on the input anymore.

Also I think the button shoudn't be shown for a `multiple` select.

ThibaudDauce:
> I find it weird that the use can't click on the input anymore.
> 
> Also I think the button shoudn't be shown for a `multiple` select.

Good catch!

nicolaskempf57 (approved):

PR #279, opened 2025-01-31
Add org certified and type

nicolaskempf57:


maudetes (approved):
:pray:
Not introduced in this PR, but I think in responsive, the cards should probably be on one column only

PR #292, opened 2025-02-05
Add FullPageTabs

nicolaskempf57:


ThibaudDauce (commented):
There is a line below the active tab which is not there in production. Do you want to add the code for the count?

ThibaudDauce (approved):

PR #298, opened 2025-02-06
New system for `currentUser`

ThibaudDauce:
- [x] Add the same as orgs for `setCurrentUser` / `currentUser`
- [x] Auto call `setCurrentUser` / `setCurrentOrganization` on API calls (when the API returns an object belonging to a user or an org)
- [x] Add all `users` to the menu (for super-admins, users fetched from API calls)
- [x] Adapt the users' "/profile" page for super-admins (for exemple API keys cannot be regenerated outside "/me" routes)
- [x] Add `AdminBreadcrumb` to auto add the namespace for the `currentUser` or the `currentOrganization` if present.
- [x] Add `BreadcrumbItem` to simplify breadcrumb creation 
- [x] Remove `default-open` for `AdminSidebarMenu`, it uses the `currentOrganization` / `currentUser` directly (so it's reactive when we navigate from page to page)


components/User/AdminUserProfilePage.vue
```diff
@@ -295,9 +297,11 @@ const loading = ref(false)
 
 const profilePicture = ref<File | null>(null)
 
+const { form } = useForm(props.user, {}, {})
 watchEffect(() => {
-  if (me.value.about === null) {
-    me.value.about = ''
+  // Why this is needed?
```
nicolaskempf57, line 302: The markdown editor doesn't accept a `null` value

nicolaskempf57 (approved):

PR #299, opened 2025-02-06
Add all illustrations in the right blue colors

agarrone:
Note : 
- Not sure to understand why there is a "img" folder and an "illustrations" folder since there is illustrations in the img folder
- Some illustrations will need to be rename to replace the ones uses currently.


nicolaskempf57:
I'm adding [an issue](https://github.com/datagouv/front-end/issues/301) to tidy up the image folders.

nicolaskempf57 (approved):

PR #304, opened 2025-02-11
Feat: add datasets search component

nicolaskempf57:
The component is only available on the design page for now


components/ActionCard/ActionCard.vue
nicolaskempf57, line 1: The tailwind classes were overridden by the heading styles

ThibaudDauce:
- [ ] Schema select doesn't work with `Error: Please set getOptionId()`
- [ ] Spatial coverage select doesn't work with `Error: Please set getOptionId()`
- [ ] License select show JSON instead of name
- [ ] On open, organization select do a "scroll jump", I think it's the problem is the image not loading in dev (not sure if it's working in production with the image). Maybe set a size of the image?
- [ ] "Reset filters" button is missing `type="button"`
- [ ] Is it useful to have a sort by select if there is only relevance?

nicolaskempf57:
> * [ ]  On open, organization select do a "scroll jump", I think it's the problem is the image not loading in dev (not sure if it's working in production with the image). Maybe set a size of the image?
There is a default size. I'm trying without the lazy, that is unnecessary here.


> * [ ]  Is it useful to have a sort by select if there is only relevance?
This is a props, I'll see if it's still necessary or add one more to have a better example

ThibaudDauce (changes requested):
When I change "Organization type" (missing translation too), I don't see a change in the results?

I cannot see "Spatial coverage" filters suggestion?

"Spatial granularity" doesn't seems to change the results either?

nicolaskempf57:
You're right, the search filters have different names.

Also I made a udata PR to avoid another request to get the geozone on page load based on the URL params : https://github.com/opendatateam/udata/pull/3267

PR #310, opened 2025-02-18
Allow routing without i18n prefix

nicolaskempf57:


nicolaskempf57:
Not necessary here, the change was needed in the reverse proxy configuration.

PR #311, opened 2025-02-19
Handle null value in contact points

nicolaskempf57:


components/Datasets/DescribeDataset.vue
```diff
@@ -346,8 +346,8 @@
           >
             <template v-for="(contact_point, index) in form.contact_points">
               <ContactPointSelect
-                v-if="contact_point.role == 'contact'"
-                :key="'id' in contact_point ? contact_point.id : index"
+                v-if="!contact_point || contact_point.role == 'contact'"
```
maudetes, line 349: Will need to be updated when merged with https://github.com/datagouv/front-end/pull/317

maudetes (approved):
What would lead to having a null contact point in the form?

I think we can't unset a contact point?

nicolaskempf57:
> What would lead to having a null contact point in the form?

When you click on the `trash` icon in the contact point select, it unsets it and sets it as null.

PR #312, opened 2025-02-20
Tailwind v4

ThibaudDauce:


assets/css/main.css
```diff
@@ -0,0 +1,105 @@
+@import "@gouvfr/dsfr/dist/dsfr.css";
+@import "@gouvfr/dsfr/dist/utility/utility.css";
+@import "@datagouv/components/dist/style.css";
+
+@import "tailwindcss" important;
```
nicolaskempf57, line 5: This change is adding a `!important` to the rule :

```
#__nuxt,
#__nuxt > div {
  @apply h-full;
}
```

This is making the `nuxt-loading-indicator` full page instead of 3px.

ThibaudDauce: Fixed by not using Tailwind for this

nicolaskempf57 (approved):

PR #316, opened 2025-02-21
Allow null end for temporal coverage

nicolaskempf57:
Part of https://github.com/datagouv/data.gouv.fr/issues/1268


maudetes (approved):
:pray:

PR #332, opened 2025-03-04
Extract and add fr translations

maudetes:


ThibaudDauce (approved):
Too bad it doesn't keep the `"": "",` at the end…

ThibaudDauce:
> > Too bad it doesn't keep the `"": "",` at the end…
> 
> We can move it to the top if you prefer I guess? :p

The goal of this line is to not have to manage the trailing comma :-) So no point in having it at the beginning of the file or in the middle :-)

PR #353, opened 2025-03-12
Add contact points

nicolaskempf57:
And some other fixes


components/AdminTable/AdminContactPointsTable/AdminContactPointsTable.vue
```diff
@@ -0,0 +1,189 @@
+<template>
+  <AdminTable>
```
maudetes, line 2: I think we're missing an empty state if there are no contact points

components/AdminTable/AdminContactPointsTable/AdminContactPointsTable.vue
```diff
@@ -0,0 +1,189 @@
...
+              <BrandedButton
+                color="secondary-softer"
+                :icon="RiPencilLine"
+                icon-only
+                size="xs"
+                keep-margins-even-without-borders
+                v-bind="attrs"
+                v-on="listeners"
+              >
+                {{ $t('Transfer') }}
```
maudetes, line 66: I think there is a wording issue here but it doesn't seem to be taken into account though. Should it be a `title` props on BrandedButton?

components/AdminTable/AdminContactPointsTable/AdminContactPointsTable.vue
```diff
@@ -0,0 +1,189 @@
...
+              <InputGroup
+                v-model="newContactForm.email"
+                type="email"
+                :label="t('Email')"
+                placeholder="contact@organization.org"
+                @blur="touch('email')"
+              />
+              <InputGroup
+                v-model="newContactForm.contact_form"
+                type="url"
```
maudetes, line 96: no front validation here (not related with this PR I think)

maudetes (approved):
:rocket:

components/AdminTable/AdminContactPointsTable/AdminContactPointsTable.vue
nicolaskempf57: I'm adding a title too for sighted user

PR #356, opened 2025-03-13
Add contact point form valdiation

nicolaskempf57:
Closes https://github.com/datagouv/cdata/issues/355


ThibaudDauce (commented):
When clicking "New contact point" then "Save" without anything I have an empty error? 
image

ThibaudDauce (approved):
Thanks for the Typescript types :-D

PR #361, opened 2025-03-18
Fix org pages

nicolaskempf57:


ThibaudDauce (approved):
Could also be done by extracting the "Ask to join…" button in a separate component if you want to avoid the manual `useAsyncData` but fine with me :-)

PR #365, opened 2025-03-18
Add support

nicolaskempf57:
This PR moves the support site to cdata.

It includes existing features like :
- cards
- Q&A
- support tree with answers
- link to a step in the anwser tree with backward-compatibility

It allows some improvments :
- support tree is now in TOML 🔥 with meaningless whitespaces
- answers don't include HTML anymore and this functionality is moved to the config file (link and label of a BrandedButton)
- When a user moves in the Q&A and the support tree, the browser URL is updated accordingly allowing an easy sharing of the current step


components/Accordion/Accordion.vue
```diff
@@ -39,16 +43,19 @@ import type { AccordionState } from '~/types/form'
 import { key, type AccordionRegister } from '~/components/Accordion/injectionKey'
 
 const props = withDefaults(defineProps<{
-  id: string | undefined
+  id?: string | undefined
```
ThibaudDauce, line 46: Why remove the optional here?

components/Accordion/Accordion.vue
```diff
@@ -79,5 +86,10 @@ const iconColor = computed(() => {
       return 'text-neutral-500'
   }
 })
-onUnmounted(() => unregister(accordionId))
+onMounted(() => {
+  if (route.hash === `#${titleAccordionId}`) {
+    open(titleAccordionId)
```
ThibaudDauce, line 91: I think it could be cool to have better ID in the hash, maybe set custom IDs for the FAQ?

components/Accordion/Accordion.vue
```diff
@@ -79,5 +86,10 @@ const iconColor = computed(() => {
...
+  }
+})
+onUnmounted(() => unregister(titleAccordionId))
```
ThibaudDauce, line 94: Why this is required?

components/Support/Choices.vue
```diff
@@ -0,0 +1,173 @@
...
+          :has-error="!!getFirstError('body')"
+          :error-text="getFirstError('body')"
+          @blur="touch('body')"
+        />
+        <div
+          v-if="messageSent"
+          class="flex w-56 text-center"
+        >
+          <SimpleBanner type="primary">
+            Message envoyé !
```
ThibaudDauce, line 62: This is not translated?

components/Support/Choices.vue
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+async function submit(question: QuestionWithSegment) {
+  if (!validate()) {
+    return
+  }
+  loading.value = true
+  try {
+    await $fetch('/api/send-message', {
```
ThibaudDauce, line 155: Could we use a different path for Nuxt API than `/api` to avoid confusion with the datagouv API?

server/api/send-message.post.ts
```diff
@@ -0,0 +1,50 @@
...
+    throw createError({
+      statusCode: 400,
+      statusMessage: 'Content type should be application/json',
+    })
+  }
+  const body = await readBody(event)
+  const CrispClient = new Crisp()
+
+  const config = useRuntimeConfig(event)
+  const websiteId = '60fe27e1-7f04-45c9-8b0d-833e95f10c79'
```
ThibaudDauce, line 15: Should it be in config?

server/api/send-message.post.ts
```diff
@@ -0,0 +1,50 @@
...
+  try {
+    const { session_id: sessionId } = await CrispClient.website.createNewConversation(websiteId)
+    await CrispClient.website.updateConversationMetas(websiteId, sessionId, {
+      email,
+      segments: ['support.data.gouv.fr', body.segment],
+    })
+    CrispClient.website.sendMessageInConversation(websiteId, sessionId, message)
+  }
+  catch (e) {
+    console.log(e)
```
ThibaudDauce, line 48: Is `console.error` better?

server/api/send-message.post.ts
```diff
@@ -0,0 +1,50 @@
...
+  }
+})
```
ThibaudDauce, line 50: Not sure about why putting this in Nuxt instead of `udata`. Maybe we could document when we use the Python API vs Nuxt API somewhere?

components/Accordion/Accordion.vue
nicolaskempf57: I'm adding it, because it was not optional until now, it was only allowing an undefined value

components/Accordion/Accordion.vue
nicolaskempf57: This opens the corresponding accordion, otherwise it is closed

components/Support/Choices.vue
nicolaskempf57: No the support page is only in french, with the corresponding `lang` attribute because the answers and configuration are only in french

components/Accordion/Accordion.vue
nicolaskempf57: I wanted to keep the addition of new accordion simple, so I relied on generated ids but we can use custom ids instead

components/Support/Choices.vue
nicolaskempf57: Indeed, it will be easier for our nginx config too. I went for api because it was the default for nuxt

server/api/send-message.post.ts
nicolaskempf57: For me, the goal of the Nuxt API is to handle things that are not linked to udata. 
That could be on the front-end only if there was no secrets involved.

components/Accordion/Accordion.vue
ThibaudDauce: Read it the other way :-)

ThibaudDauce (approved):

PR #367, opened 2025-03-19
Improve discussions

ThibaudDauce:
- [x] Merge the two implementations of discussions table
- [x] Add filter on discussion status
- [x] Add respond and close modal


components/AdminTable/AdminDiscussionsTable/AdminDiscussionsTable.vue
```diff
@@ -21,18 +28,26 @@
...
           @sort="(direction: SortDirection) => $emit('sort', 'closed', direction)"
         >
           {{ t("Closed at") }}
         </AdminTableTh>
         <AdminTableTh
           scope="col"
-          class="w-1/5"
         >
-          {{ t("Link to the discussion") }}
+          {{ t("Actions") }}
```
nicolaskempf57, line 50: Can we force a width to allow both buttons to be on the same line ?

components/AdminTable/AdminDiscussionsTable/AdminDiscussionsTable.vue
```diff
@@ -83,47 +106,58 @@
...
             {{ formatDate(discussion.closed) }}
           </template>
         </td>
         <td>
-          <a
-            v-if="subjects[discussion.subject.id]"
-            class="fr-link inline-flex"
-            :href="getDiscussionUrl(discussion.id, subjects[discussion.subject.id])"
-            target="_blank"
+          <BrandedButton
```
nicolaskempf57, line 118: Titles are necessary to make the button accessibles

nicolaskempf57 (approved):
I think that implementing the discussions in cData is becoming a priority with this new modal

components/AdminTable/AdminDiscussionsTable/AdminDiscussionsTable.vue
ThibaudDauce: The button already has a content? `icon-only` props add a `sr-only` on the text to not show it but screen reader should be able to read it?

components/AdminTable/AdminDiscussionsTable/AdminDiscussionsTable.vue
ThibaudDauce: Done in 731839a

PR #369, opened 2025-03-19
Use APIv2 for datasets

ThibaudDauce:
Require https://github.com/opendatateam/udata/pull/3290


components/DatasetsSelect.vue
```diff
@@ -103,7 +103,7 @@ import SearchableSelect from '~/components/SearchableSelect.vue'
 import type { DatasetSuggest } from '~/types/types'
 
 const sortableRootRef = useTemplateRef('sortableRoot')
-const datasetsById = ref<Record<string, Dataset>>({})
+const datasetsById = ref<Record<string, Dataset | DatasetV2>>({})
```
nicolaskempf57, line 106: Can we change the type to DatasetV2 only ?

nicolaskempf57 (approved):
Can we do it for the `AdminDatasetsTable` too ?

PR #376, opened 2025-03-27
Fix catalog download

nicolaskempf57:


maudetes (approved):
Thank you for the fix!

PR #384, opened 2025-04-09
Add pages

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1682


ThibaudDauce:
Do you also have black text in the buttons?

image

components/MarkdownEditor/InternalEditor.vue
```diff
@@ -90,7 +90,7 @@
     </div>
     <div class="p-1">
       <Milkdown
-        :class="markdownClasses"
+        :class="markdownSmClasses"
```
ThibaudDauce, line 93: Could we add explicit imports for this? If we migrate this component to datagouv/components it will stop working in `udata-front-kit`

ThibaudDauce (approved):

nicolaskempf57:
> Do you also have black text in the buttons?
> 
> image

Buttons parent should have a `not-prose` now to avoid this. But I suppose I missed this one. Where it is located ?

PR #397, opened 2025-04-23
Add noindex & nofollow meta in admin

maudetes:
All pages located in admin require authentication, so this is probably useless.
However, it was present on the old udata admin so I'm adding it here as an additional safety measure.

What do you think?


ThibaudDauce:
Not a big fan since it's behind authentification it shouldn't be required…

maudetes:
Let's not add it then :)

PR #398, opened 2025-04-24
Add reuses

nicolaskempf57:


ThibaudDauce (commented):
- « See reuse » is not translated?
- When a click « Discussions », I get a 404
- Small problem with dataset's cards (but maybe it's my broken CSS/images?)
image
- « 5 réutilisations du même créateur » contains the current reuse too.

nicolaskempf57:
> * « See reuse » is not translated?

✅ 

> * When a click « Discussions », I get a 404

I'll try to reproduce, was it with a reuse from dev ?

> * Small problem with dataset's cards (but maybe it's my broken CSS/images?)

I'll try to reproduce for the issue with the acronym

> * « 5 réutilisations du même créateur » contains the current reuse too.

This is strange because there is an explicit front-end filter based on the current reuse id 🤔

ThibaudDauce:
> > * « See reuse » is not translated?
> 
> ✅
> 
> > * When a click « Discussions », I get a 404
> 
> I'll try to reproduce, was it with a reuse from dev ?
> 
> > * Small problem with dataset's cards (but maybe it's my broken CSS/images?)
> 
> I'll try to reproduce for the issue with the acronym
> 
> > * « 5 réutilisations du même créateur » contains the current reuse too.
> 
> This is strange because there is an explicit front-end filter based on the current reuse id 🤔

I think I wasn't up to date… Oups :-(

Discussions is ok and the filter for same reuse is ok too.

I still have the first dataset new line problem (but I have it too on my cards in the dataservice page). For exemple on http://dev.local:3000/fr/reuses/estimation-de-prix-immobilier-en-ligne/. It's not the acronym because the second one has an acronym but no problem…
image

nicolaskempf57:
> I still have the first dataset new line problem (but I have it too on my cards in the dataservice page). For exemple on http://dev.local:3000/fr/reuses/estimation-de-prix-immobilier-en-ligne/. It's not the acronym because the second one has an acronym but no problem…

It should be better now 🙏🏻

components/FollowButton.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <button
```
ThibaudDauce, line 2: Why not use `BrandedButton`, buttons doesn't have the same size between favorite, modify and see the ruese

ThibaudDauce (approved):

components/FollowButton.vue
nicolaskempf57: I just find out that we have a `iconAttrs` props to handle this case ! I'm changing it

nicolaskempf57:
FYI, @ThibaudDauce I updated the `<BrandedButton>` so now the component handles the external icon, instead of relying on the DSFR implementation. It allows us to apply the same icon size logic to this new one too.

PR #405, opened 2025-05-05
Use query params instead of hash segments for resources edit routes

ThibaudDauce:
It allows to redirect to the route from Nginx (which doesn't see hashes)


maudetes (approved):
:pray:

PR #417, opened 2025-05-12
Fix enter and miss clicks in search menu

ThibaudDauce:
I'm not a big fan of removing the anchor tag but didn't find another solution…

### DIFF
--- a/components/MenuSearch/MenuSearch.vue
+++ b/components/MenuSearch/MenuSearch.vue
@@ -1,5 +1,6 @@
 <template>
   <Combobox
+    v-model="selectedItem"
     as="div"
     class="relative"
   >
@@ -36,16 +37,13 @@
           :key="item.type"
           v-slot="{ active }"
           as="template"
+          :value="item"
         >
           <li
             class="relative cursor-default select-none px-4 hover:bg-gray-some *:last:border-0"
             :class="{ 'text-datagouv': active }"
           >
-            <NuxtLink
-              :to="item.to"
-              :external="item.external"
-              class="flex items-center space-x-2 border-b py-3"
-            >
+            <div class="flex items-center space-x-2 border-b py-3">
               <component
                 :is="item.icon"
                 class="h-4 w-4"
@@ -78,7 +76,7 @@
               <div aria-hidden="true">
                 <RiArrowRightSLine class="h-4 w-4" />
               </div>
-            </NuxtLink>
+            </div>
           </li>
         </ComboboxOption>
       </ComboboxOptions>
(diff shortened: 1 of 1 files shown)


components/MenuSearch/MenuSearch.vue
```diff
@@ -89,10 +87,24 @@
...
+  to: string
+  external: boolean
+}
 
 const { t } = useI18n()
 const localePath = useLocalePath()
 const query = ref('')
+const selectedItem = ref<null | Item>(null)
+
+watch(selectedItem, () => {
```
maudetes, line 104: I think we also navigate if we click out of the menu?

ThibaudDauce: I don't reproduce the problem… Are you sure?

ThibaudDauce: Seems to be a bug with Headless UI https://github.com/tailwindlabs/headlessui/issues/3615 but trying to fix it…

ThibaudDauce: Loved the one hour investigation for the one line fix :-) 0e1ab48

components/MenuSearch/MenuSearch.vue
```diff
@@ -36,16 +43,13 @@
           :key="item.type"
           v-slot="{ active }"
           as="template"
+          :value="item"
         >
           <li
```
maudetes, line 48: Not related to this PR, but I think the `aria-selected` is always false?

maudetes (commented):
On my side, seems to work nicely thank you! :rocket:

Additionally, I'd like @nicolaskempf57 to take a look for an accessibility opinion on this fix?

nicolaskempf57:
> On my side, seems to work nicely thank you! 🚀
> 
> Additionally, I'd like @nicolaskempf57 to take a look for an accessibility opinion on this fix?

There is one keyboard interaction that seems odd.

When you are focusing the input, you can type something, it will then show the list below, selected on the dataset page search. 

This is expected BUT if you tab outside of the component, it moves you to the dataset page instead of allowing you to leave the combobox.

ThibaudDauce:
> > On my side, seems to work nicely thank you! 🚀
> > Additionally, I'd like @nicolaskempf57 to take a look for an accessibility opinion on this fix?
> 
> There is one keyboard interaction that seems odd.
> 
> When you are focusing the input, you can type something, it will then show the list below, selected on the dataset page search.
> 
> This is expected BUT if you tab outside of the component, it moves you to the dataset page instead of allowing you to leave the combobox.

This is the expected behavior of the Combobox from headless UI. See Tab key in https://headlessui.com/v1/vue/combobox#keyboard-interaction

nicolaskempf57:
> This is the expected behavior of the Combobox from headless UI. See Tab key in https://headlessui.com/v1/vue/combobox#keyboard-interaction

I think it's fair for a real combobox but it seems strange now that we fixed this behavior for the click and not for the tab

ThibaudDauce:
> > This is the expected behavior of the Combobox from headless UI. See Tab key in https://headlessui.com/v1/vue/combobox#keyboard-interaction
> 
> I think it's fair for a real combobox but it seems strange now that we fixed this behavior for the click and not for the tab

I think the click outside is supposed to be a cancel and a tab is suppose to be a select in accessibility but I'm not sure

nicolaskempf57 (approved):

PR #419, opened 2025-05-13
Add feature to reuse administration

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1723

### DIFF
--- a/components/Reuses/AdminUpdateReusePage.vue
+++ b/components/Reuses/AdminUpdateReusePage.vue
@@ -148,7 +148,12 @@ async function feature() {
       method,
     })
     await refresh()
-    toast.success(t('Réutilisation mise en avant !'))
+    if (method === 'DELETE') {
+      toast.success(t('Réutilisation retirée de la mise en avant !'))
+    }
+    else {
+      toast.success(t('Réutilisation mise en avant !'))
+    }
   }
   catch {
     toast.error(t('Impossible de mettre en avant cette réutilisation'))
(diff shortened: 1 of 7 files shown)


components/Reuses/AdminUpdateReusePage.vue
```diff
@@ -135,7 +136,25 @@ async function deleteReuse() {
...
+
+async function feature() {
+  const method = reuse.value.featured ? 'DELETE' : 'POST'
+  try {
+    start()
+    await $api(`/api/1/reuses/${route.params.id}/featured`, {
+      method,
+    })
+    await refresh()
+    toast.success(t('Réutilisation mise en avant !'))
```
ThibaudDauce, line 151: Should be not the same depending on the value?

ThibaudDauce (approved):
Ok for me. But I think we should refactor the design of these booleans :
- `featured` is a checkmark with auto-save on click
- `draft` is a checkbox with no auto-save
- `delete` is a banner

In my opinion all these "actions" may be banners with buttons. I don't know if @agarrone have an idea?

nicolaskempf57:
> Ok for me. But I think we should refactor the design of these booleans :
> 
> * `featured` is a checkmark with auto-save on click
> * `draft` is a checkbox with no auto-save
> * `delete` is a banner
> 
> In my opinion all these "actions" may be banners with buttons. I don't know if @agarrone have an idea?

I think transfer and delete are important actions and I think it's fair that they are handled in a banner.

Also I think the banner component works well only if there is only 1 or 2 of them.

With a list of 3 or 4, there is a need to differentiate them based on relative importance.

PR #426, opened 2025-05-14
Add missing trailing slashes in discussions API calls

ThibaudDauce:
… and add specific errors' messages for 429 and 403


maudetes (approved):
:pray:

PR #427, opened 2025-05-14
Fix double queries on load

ThibaudDauce:
After the first API call, we set the org ID as the current one, which change the currentOrgId which is used in the params for the discussions API, so even if it's the same ID the computed is rerun, the params object change and Nuxt redo the API call.


maudetes (approved):
:pray: thank you for the fix!

PR #431, opened 2025-05-16
Add discussion to admin

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1692
This PR adds the new discussion thread to the admin response modale.

It also fixes modal without footer and modal scroll.

It prevents click to close. 
We might want to block it only for fullscreen modal, what do you think ?


ThibaudDauce (approved):
I don't see the fullscreen modal being used for discussion, is it wanted?

PR #441, opened 2025-05-20
Fix removing last contact point

ThibaudDauce:


ThomasG77:
Same issue on dataservices.ts side while checking if same issue elsewhere in basecode e.g https://github.com/datagouv/cdata/blob/main/utils/dataservices.ts#L38
Fix here or open a new PR. Up to you

nicolaskempf57 (approved):

PR #447, opened 2025-05-22
Prevent test in production

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1727

### DIFF
--- a/components/Datasets/DescribeResource.vue
+++ b/components/Datasets/DescribeResource.vue
@@ -48,43 +48,45 @@
         v-if="form.filetype === 'file' || !form.filetype"
         form-key="file"
       >
-        <div v-if="newFile">
-          <label
-            v-if="type === 'update'"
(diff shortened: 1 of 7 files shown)


components/Datasets/DescribeResource.vue
```diff
@@ -355,8 +361,8 @@ const setFiles = (files: Array<File>) => {
   }
 }
 
-const { data: extensions } = await useAPI<Array<string>>('/api/1/datasets/extensions/')
-const { data: schemas } = await useAPI<SchemaResponseData>('/api/1/datasets/schemas/')
+const { data: extensions, status } = await useAPI<Array<string>>('/api/1/datasets/extensions/', { lazy: true })
```
nicolaskempf57, line 364: These changes and the new loading blocks prevents the modal showing with an empty body while the component is loading async data

PR #448, opened 2025-05-22
Add button to delete user as spam

maudetes:
It existed previously in udata-front (https://github.com/datagouv/udata-front/pull/666)


components/User/AdminUserProfilePage.vue
```diff
@@ -255,7 +255,7 @@
 
         <template #button>
           <ModalWithButton
-            :title="$t('Are you sure you want to delete this organization ?')"
+            :title="$t('Are you sure you want to delete this user ?')"
```
ThibaudDauce, line 258: :fr: ?

ThibaudDauce (approved):

components/User/AdminUserProfilePage.vue
maudetes: I didn't want to mix language :sweat_smile:

PR #455, opened 2025-06-02
chore: migrate to topics elements

abulte:
Handles the impact of Topic.elements introduced in https://github.com/opendatateam/udata/pull/3318.

The lists of datasets and reuses for a Topic are now read-only. We would need to properly handle elements instead of generic lists for write operations. I think this can be done later? Not sure if this feature is widely used.


components/AdminTable/AdminReusesTable/AdminReusesTable.vue
```diff
@@ -120,12 +120,14 @@ import AdminContentWithTooltip from '../../../components/AdminContentWithTooltip
...
-const props = defineProps<{
+const props = withDefaults(defineProps<{
   activities: Record<string, Activity>
   reuses: Array<Reuse>
   sortedBy: ReuseSortedBy
   sortDirection: SortDirection
-}>()
+}>(), {
+  activities: () => ({}),
+})
```
maudetes, line 130: I had some issue with missing activities in topics/[id]/reuses.vue. Fixed it in https://github.com/datagouv/cdata/pull/455/commits/51bc264ace3a887aa28b0fd63dbdd6edc0ed4207

pages/admin/topics/[id]/index.vue
```diff
@@ -54,7 +54,7 @@ const { t } = useI18n()
 
 const save = async () => {
   if (await validate()) {
-    await $api(`/api/1/topics/${props.topic.id}/`, {
+    await $api(`/api/2/topics/${props.topic.id}/`, {
```
Comment (maudetes) [inline pages/admin/topics/[id]/index.vue:57]: I think we should PUT on topic with api v2. Done in https://github.com/datagouv/cdata/pull/455/commits/b63a50a27fbd7d555fa5e29bb0d3a5b986eb7223

PR #458, opened 2025-06-03
Remove some externals

ThibaudDauce:


pages/admin/datasets/new.vue
```diff
@@ -178,7 +178,7 @@ async function updateDataset(asPrivate: boolean) {
     }
   }
 
-  await navigateTo(newDataset.value.page, { external: true })
+  await navigateTo(newDataset.value.page)
```
nicolaskempf57, line 181: Using `.page` without external will break on local

nicolaskempf57 (approved):
I don't know if we prefer to break local or reduce performance for our deployed env

pages/admin/datasets/new.vue
ThibaudDauce: I just tested and the redirects seems to work, even to localhost:7000 and without the external

PR #479, opened 2025-06-11
Add commit version in footer

ThibaudDauce:


components/SiteFooter/SiteFooter.vue
```diff
@@ -188,11 +188,11 @@
...
                 <li>
                   <a
-                    href="https://github.com/datagouv/cdata/"
-                    :title="$t('data.gouv.fr extension: cdata')"
+                    :href="config.public.commitId ? `https://github.com/datagouv/cdata/commit/${config.public.commitId}` : 'https://github.com/datagouv/cdata/'"
+                    :title="config.public.commitId ? $t('Version {version}', { version: config.public.commitId }) : undefined"
                     class="fr-footer__top-link"
                   >
-                    {{ $t('data.gouv.fr front-end') }}
+                    {{ $t('Interface utilisateur de data.gouv.fr : cdata') }}
```
maudetes, line 195: I think showing the version explicitly in the text would be nicer to our re-users (and not sure adding it in title is a11y?)

maudetes (approved):
Nice, we can go with this!

I think in a second iteration, we could use confs set in our infra repo instead (since we could usage commits, tags, etc.), to also have the udata version

PR #485, opened 2025-06-12
Fix edit resource link

ThibaudDauce:


ThibaudDauce:
> The `ResourceAccordion` in `pages/datasets/[did]/community-resources.vue` file is missing both `isCommunityResource` and `canEdit` props.
> 
> Also the resource `canEdit` is only based on being an admin which seems wrong.

It's an error if we don't have the `isCommunityResource` in this page… Need to fix this.

And concerning the `canEdit` it's wrong in a lot of places ("Modify" buttons in page headers too). We currently only check for `isAdmin()` because we can only do this. I've planned to make a PR on udata on monday to add the permissions objects to all models so we can use it in cdata to show the buttons to orgs owners too.

nicolaskempf57 (approved):

PR #503, opened 2025-06-19
Use dedicated analysis extras to know if a resource has a preview

maudetes:
It should not be merged before https://github.com/datagouv/hydra/pull/279 has been deployed in production and a migration has been run on stock


ThibaudDauce:
Done in https://github.com/datagouv/cdata/pull/554

PR #510, opened 2025-06-23
Add restore banner and object status

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1750

### DIFF
--- a/components/Dataservices/AdminUpdateDataservicePage.vue
+++ b/components/Dataservices/AdminUpdateDataservicePage.vue
@@ -9,6 +9,7 @@
     >
       <template #top>
         <BannerAction
+          v-if="!dataservice.deleted_at && !dataservice.archived_at"
           class="mb-4"
           type="primary"
           :title="$t(`Modifier la visibilité de l'API`)"
--- a/components/Datasets/AdminUpdateDatasetPage.vue
+++ b/components/Datasets/AdminUpdateDatasetPage.vue
@@ -11,6 +11,7 @@
     >
       <template #top>
         <BannerAction
+          v-if="!dataset.deleted && !dataset.archived"
           class="mb-4"
           type="primary"
           :title="$t('Modifier la visibilité du jeu de données')"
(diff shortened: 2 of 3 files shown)


components/Dataservices/AdminUpdateDataservicePage.vue
```diff
@@ -7,6 +7,26 @@
       type="update"
       @submit="save"
     >
+      <template #top>
+        <BannerAction
+          v-if="dataservice.deleted_at"
+          class="mb-4"
+          type="warning"
+          :title="$t('Restaurer cette API')"
+        >
```
ThibaudDauce, line 16: Should be only for global admins?

components/Datasets/AdminUpdateDatasetPage.vue
```diff
@@ -9,6 +9,26 @@
       @feature="feature"
       @submit="save"
     >
+      <template #top>
+        <BannerAction
+          v-if="dataset.deleted"
+          class="mb-4"
+          type="warning"
+          :title="$t('Restaurer ce jeu de données')"
```
ThibaudDauce, line 17: Should be only for global admins?

components/Reuses/AdminUpdateReusePage.vue
```diff
@@ -7,6 +7,26 @@
       @feature="feature"
       @submit="save"
     >
+      <template #top>
+        <BannerAction
+          v-if="reuse.deleted"
+          class="mb-4"
+          type="warning"
+          :title="$t('Restaurer la réutilisation')"
```
ThibaudDauce, line 15: Should be only for global admins?

ThibaudDauce (approved):
Working nice, I just think we've decided to allow only admins to restore objects from deletion?

ThibaudDauce:
> > Working nice, I just think we've decided to allow only admins to restore objects from deletion?
> 
> We don't want organizations to be restored, but I think we want it for other objects

Ok if it's what @agarrone wants :-)

agarrone:
> > > Working nice, I just think we've decided to allow only admins to restore objects from deletion?
> > 
> > 
> > We don't want organizations to be restored, but I think we want it for other objects
> 
> Ok if it's what @agarrone wants :-)

Yes I think it's the good behaviour. We'll see if it causes problems.

PR #521, opened 2025-06-25
Fix updating wrong avatar for global admins

ThibaudDauce:


pages/admin/users/[uid]/profile.vue
```diff
@@ -2,11 +2,25 @@
...
+  // const newUser = await $api<User>(`/api/1/users/${user.value.id}/`)
+
+  // This doesn't work I don't know why…
+  // users.value[user.value.id] = newUser
+
+  // This doesn't work because of useComponentConfig outside @datagouv/components in
+  // <Avatar />
+  // user.value.avatar = newUser.avatar
+  // user.value.avatar_thumbnail = newUser.avatar_thumbnail
+}
```
Comment (maudetes) [inline pages/admin/users/[uid]/profile.vue:25]: Do we want to keep this?

PR #559, opened 2025-07-08
Replace NuxtLinkLocale by smart CdataLink to manage absolute URL

ThibaudDauce:


components/Harvesters/AdminHarvestersPage.vue
```diff
@@ -85,7 +85,7 @@
             >
               <td>
                 <AdminContentWithTooltip>
-                  <NuxtLinkLocale
```
nicolaskempf57, line 88: There is no href fo these links so they are disabled

ThibaudDauce: Fixed in b07c1b6

PR #575, opened 2025-07-16
Fix switch org tab

ThibaudDauce:


maudetes:
I'm not sure, but now going on an organization we're not a member of (as super-admin) does not add this organization in the sidebar anymore?

PR #589, opened 2025-07-16
fix: load preview if remote, whatever the CORS issue, and display err…

bolinocroustibat:
…or if CORS issue


maudetes (commented):
Nice! Thank you for this iteration, I have some questions on the cors detection

maudetes (approved):
Thank you! Indeed, we should find something better but it would still be nice to have this already :sparkles:

PR #613, opened 2025-07-25
Use same logic for all 3 searches

nicolaskempf57:


ThibaudDauce (approved):
It could be nice to have a composable about this… But not require for this change

PR #618, opened 2025-07-29
Try to reduce build time by moving resource components to client and async

nicolaskempf57:


maudetes (approved):
Seems to work nicely, thank you (from ~13min to 8min locally for me)!

PR #620, opened 2025-07-30
Add max-w to post page

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1823

### DIFF
--- a/pages/posts/[id].vue
+++ b/pages/posts/[id].vue
@@ -34,7 +34,7 @@
       <template v-if="post.published || isMeAdmin()">
         <p
           v-if="post.published"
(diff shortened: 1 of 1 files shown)


ThibaudDauce (approved):
I don't like the centering (I would prefer left align) but I don't know what @agarrone thinks?

nicolaskempf57:
> I asked for a smaller container for articles. I believe it enhances readability especially on larger screens. But the text should be left align anyway, do you have a screenshot ? 🙂

The text is left-aligned but the container is centered.

Xnapper-2025-07-31-09 14 22

PR #643, opened 2025-08-25
Add change email flash message

maudetes:
I did not read the target email value from flash_data since it is poorly encoded and not valid JSON at the moment.


ThibaudDauce (approved):
Can you mark this as temporary? We'll don't need this when security pages are merged in cdata.

PR #651, opened 2025-08-27
Update logos

ThibaudDauce:
- Fix https://github.com/datagouv/data.gouv.fr/issues/1767


public/nuxt_images/favicon.png
maudetes, line 1: Just for you information, the background is transparent even between the cube parts. It shows like this on my firefox theme
image

nuxt.config.ts
```diff
@@ -22,10 +22,7 @@ export default defineNuxtConfig({
         class: 'datagouv-components',
       },
       link: [
-        // Cannot use `/public/favicon.png` because the reverse proxy is calling udata-front for `/`
-        // When the migration is over we can remove this static path.
-        // :AfterMigration
-        { rel: 'shortcut icon', href: 'https://static.data.gouv.fr/_themes/gouvfr/img/favicon.png' },
+        { rel: 'shortcut icon', href: '/nuxt_images/favicon.png' },
```
maudetes, line 25: Can't we just add the acl rule? I would rather target the final URL

components/AdminPublishButton/AdminPublishButton.vue
```diff
@@ -6,14 +6,15 @@
     :external="type === 'harvester'"
   >
     <span v-if="type === 'harvester'">{{ t("Créer un moissoneur sur") }}</span>
-    <span v-else>{{ t("Publier sur") }}</span>&nbsp;<SiteLogo />
+    <span v-else>{{ t("Publier sur") }}</span>&nbsp;<LogoAsText />
```
maudetes, line 9: I think using the text only from the new logo would do?
image

maudetes (approved):

public/nuxt_images/favicon.png
ThibaudDauce: I thinks it's wanted?

nuxt.config.ts
ThibaudDauce: This is the final URL :-) I think we'll keep the `nuxt_images` folder for the time being. When we flip the switch to route everything to cdata except the API we may rename it to just `images` but it's a quick change.

PR #657, opened 2025-08-28
Do not send form on search

ThibaudDauce:
Since the search is live (without refresh) we don't need to send the form on clic.


maudetes (commented):
I think it should still trigger a search with the loading status and then result being displayed. Else, the lack of feedback makes it feel like the search is broken.

PR #665, opened 2025-09-02
Use last activity in object table

maudetes:
Fix incoherent activity being shown in AdminObjectTable `Mis à jour le`.
The `getActitiesForObjects` would fetch a page of 20 activities by object and build the activities by object dict on it, resulting in the latest activity of the page to be mapped.

For example, on the [admin datasets page](https://www.data.gouv.fr/admin/organizations/646b7187b50b2a93b1ae3d45/datasets), we have 
image
VS [on the dataset itself](https://www.data.gouv.fr/admin/datasets/6889cc10e79a25f17ed69acb)
image

Use a default page_size of 1 in getActitiesForObjects to make sure to get the first activity, in our sorted on -created_at.

However, the admin column `Mist à jour le` based on activity does not equal the information shown in the public page of `Dernière mise à jour` that is based on `dataset.last_update` and which will differ in the case of harvested dataset remote resources.
Should we rename the column (example: `Mis à jour sur data.gouv.fr le`)? Any other idea?

### DIFF
--- a/components/Dataservices/AdminDataservicesPage.vue
+++ b/components/Dataservices/AdminDataservicesPage.vue
@@ -141,7 +141,7 @@ const { data: pageData, status, refresh } = await useAPI<PaginatedArray<Dataserv
 
 watchEffect(async () => {
   if (pageData.value) {
-    const activities = await getActitiesForObjects($api, pageData.value.data)
+    const activities = await getLatestActivitiesForObjects($api, pageData.value.data)
     dataserviceActivities.value = { ...dataserviceActivities.value, ...activities }
   }
 })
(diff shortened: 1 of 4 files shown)


nicolaskempf57:
> Use a default page_size of 1 in getActitiesForObjects to make sure to get the first activity, in our sorted on -created_at.
> 
> However, the admin column `Mist à jour le` based on activity does not equal the information shown in the public page of `Dernière mise à jour` that is based on `dataset.last_update` and which will differ in the case of harvested dataset remote resources. Should we rename the column (example: `Mis à jour sur data.gouv.fr le`, `Dernière activité sur data.gouv.fr le`)? Any other idea?

I think we should call it `Dernière activité sur data.gouv.fr le` to use a diffent label as the public page

maudetes:
> we should call it `Dernière activité sur data.gouv.fr le` to use a diffent label as the public page

Done in https://github.com/datagouv/cdata/pull/665/commits/54c8a7aebaa724102c18501fa4f177fbc32a9f87

PR #673, opened 2025-09-05
Simplify README

nicolaskempf57:


maudetes (approved):
Thank you! Maybe if needed, we can move the instructions in troubleshoot, but seems like we don't need to (pre-)build datagouv-components directly anymore

PR #686, opened 2025-09-16
Fix matomo client navigation

nicolaskempf57:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1841

The code is taken in part from https://github.com/AmazingDreams/vue-matomo

The package vue-matomo isn't used because no new version was release since 3 years.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,16 @@ If you encounter `EBADF` errors, you may be affected by [this chokidar issue](ht
    }
    ```
 
+### Analytics
+
+cdata uses Matomo to track page visits. It can be disable on a per-page basis with the meta `matomoIgnore`.
+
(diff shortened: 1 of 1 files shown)


plugins/matomo.client.ts
maudetes, line 1: I have a question related on this PR, but how do we deactivate matomo script on admin & security pages?

maudetes (approved):
I trust you on this implementation!

Seems like there are different issues on [implementation with Nuxt](https://github.com/AmazingDreams/vue-matomo/issues?q=is%3Aissue%20nuxt)!
We can now go with our implementation for now

maudetes (approved):
:ok_hand:

PR #688, opened 2025-09-16
Extract SelectAudiencesTypes to component

ThibaudDauce:


pages/datasets/[did].vue
```diff
@@ -329,11 +337,60 @@
...
+                  type="pink"
+                >
+                  {{ $t('Accès restreint') }}
+                </AdminBadge>
+              </div>
+              <p class="font-bold text-xl">
+                {{ $t('Ces données ne sont accessibles que sur habilitation') }}
+              </p>
+              <p
+                v-if="dataset.access_type_reason"
```
Comment (nicolaskempf57) [inline pages/datasets/[did].vue:362]: Do we want to show the reason category if we only have it ?

Comment (ThibaudDauce) [reply pages/datasets/[did].vue]: Fixed in f4ef0de

PR #692, opened 2025-09-23
feat: ressource deletion confirmation modal wording

bolinocroustibat:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1861

screenshot 2025-09-23 at 12 40 27

screenshot 2025-09-23 at 12 50 39

### DIFF
--- a/components/Datasets/FileEditModal.vue
+++ b/components/Datasets/FileEditModal.vue
@@ -48,7 +48,7 @@
         <div class="w-full flex gap-4">
           <!-- Confirmation modal only if file has changed -->
           <ModalWithButton
-            v-if="hasFileChanged"
+            v-if="hasTabularApi && hasFileChanged"
             :title="$t('Êtes-vous sûr de vouloir modifier cette ressource ?')"
             size="lg"
           >


components/Datasets/FileEditModal.vue
```diff
@@ -45,14 +45,52 @@
     <template #footer="{ close }">
       <div class="w-full">
         <div class="w-full flex gap-4">
-          <BrandedButton
-            color="primary"
-            type="submit"
-            :form="formId"
-            :loading
+          <ModalWithButton
```
nicolaskempf57, line 48: I think this dialog is quite confusing as a replacement of the submit button and requires some conditions.

It should be shown only when necessary, which is only when someone change the file after the crawler passed on the previous one. 
I don't think we should show it inside the creation of a new dataset or on metadata changes.

components/Datasets/FileEditModal.vue
```diff
@@ -45,14 +45,52 @@
...
+                :loading
+                v-bind="attrs"
+                v-on="listeners"
+              >
+                {{ t('Valider') }}
+              </BrandedButton>
+            </template>
+            <div class="fr-mt-4w">
+              <p>
+                {{ $t('Attention : l\'') }}
```
nicolaskempf57, line 65:
```suggestion
                {{ $t(`Attention : l'`) }} 
```

components/Datasets/FileEditModal.vue
```diff
@@ -86,8 +124,32 @@
...
+                    {{ $t('fournie par data.gouv.fr.') }}
+                  </p>
+                  <p>
+                    {{ $t('Si vous supprimez la ressource,') }} 
+                    <a 
+                      :href="`https://tabular-api.data.gouv.fr/api/resources/${resource.resource?.id}/`"
+                      target="_blank"
+                      class="fr-link"
+                    >
+                      {{ $t('l\'API') }}
```
nicolaskempf57, line 148:
```suggestion
                      {{ $t(`l'API`) }}
```

components/Datasets/FileEditModal.vue
bolinocroustibat: You're right, thanks, I'll have a look into that!

components/Datasets/FileEditModal.vue
```diff
@@ -144,9 +234,73 @@ const route = useRoute()
...
+    filesize: props.resource.resource?.filesize,
+    title: props.resource.resource?.title
+  })
+  
+  // Debug: Log the complete originalResource structure
+  console.log('FileEditModal - Complete originalResource:', props.resource.resource)
+  console.log('FileEditModal - originalResource JSON:', JSON.stringify(props.resource.resource, null, 2))
+  
+  // Function to check if file has changed based on metadata
+  const checkFileChange = () => {
```
nicolaskempf57, line 259: I think this is better ! Somehow, when I change a file, it never goes to `checkFileChange`. Is it working on your side ?
Also it's not checking for changes for a remote resource.

components/Datasets/FileEditModal.vue
```diff
@@ -86,8 +152,32 @@
                 </BrandedButton>
               </template>
               <p class="fr-text--bold">
-                {{ $t("Cette action est irréversible.") }}
+                {{ $t('Cette action est irréversible.') }}
               </p>
+                <div class="fr-mt-4w">
+                  <p>
+                    {{ $t('Attention : cette ressource est exposée via une') }}
```
nicolaskempf57, line 159: I think we have quite the same issue for the delete dialog.

We should show this message only for file "apified".

bolinocroustibat: Stupid me, we can do it using `props.resource.resource?.extras?.['analysis:parsing:parsing_table']` I think. Done

PR #696, opened 2025-09-24
Basic new translation system

ThibaudDauce:


app.vue
```diff
@@ -44,10 +44,8 @@ app.vueApp.use(datagouv, {
 
 useHeadSafe({
   htmlAttrs: {
-    lang: i18nHead.value.htmlAttrs!.lang,
+    lang: locale,
   },
-  link: [...(i18nHead.value.link || [])],
-  meta: [...(i18nHead.value.meta || [])],
```
nicolaskempf57, line 50: This one was adding alternative pages for SEO purpose, for multilingual website this is required but it has no usage for us with only one language (more on the subject on https://developers.google.com/search/docs/specialty/international/localized-versions?hl=fr)

ThibaudDauce:
> > * [x]   Remove some code that called `localePath()` (to build with prefix?) or `localeRoute()` to get the full route object.
> 
> I think this one and the SEO one will prevent cdata to handle multiple languages at the same time. This is not an issue for us but it is for other udata users. I think we should talk about the stance we want to adopt regarding cdata or at revalidate or previous talks about it.

I don't think we should support multiple language via URL prefix. We can support multiple language via `Accept-Language` header and it's good enough?

nicolaskempf57:
> I don't think we should support multiple language via URL prefix. We can support multiple language via `Accept-Language` header and it's good enough?

I don't know if it's enough for the legal requirements of multilingual countries, but it's not enough for Google to index multiple versions of the site.

[Their page on the subject](https://developers.google.com/search/docs/specialty/international/managing-multi-regional-sites?hl=fr#use-different-urls-for-different-language-versions) is quite clear that :
1. each lang should have localized urls
2. each version should be known for SEO purpose (with hreflang and/or sitemap)
3. we shouldn't redirect users from one lang to another automatically

PR #699, opened 2025-09-26
Add metrics for user

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1853


components/Metrics/AdminReusesPage.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+  <div>
+    <div class="flex justify-end -mt-14 pt-0.5 mb-5">
+      <BrandedButton
+        color="secondary"
+        :disabled="!downloadStatsUrl"
```
ThibaudDauce, line 6: Maybe do not show on `null`?

components/Metrics/AdminReusesPage.vue
```diff
@@ -0,0 +1,199 @@
...
+const page = ref(1)
+const pageSize = ref(20)
+const sortedBy = ref<ReuseSortedBy>('created')
+const direction = ref<SortDirection>('desc')
+const sortDirection = computed(() => `${direction.value === 'asc' ? '' : '-'}${sortedBy.value}`)
+
+const params = computed(() => {
+  return {
+    organization: props.organization?.id,
+    user: props.user?.id,
```
ThibaudDauce, line 174: I don't understand why but I see all the reuses of the application… Same with dataservices. Datasets are working… Maybe a problem with the new api fields in udata?

nicolaskempf57: I think the search are missing the `user`, I'll make a PR on the udata  side

PR #700, opened 2025-09-26
Fix geozone not filled

nicolaskempf57:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1860

The dataset API sends us the zone id only and we need the full object for the filter (to show the name). The previous implementation was matching the zone id with an empty array (`[]`) so nothing was shown.

The new implementation send a request to `/spatial/zone/:id` before transforming the dataset API result to the dataset form.


ThibaudDauce (commented):
The dataset API has a special flag to return the full objects (and avoid doing too much work on the frontend. Have you try this?

```
const { data: dataset, status } = await useAPI<DatasetV2WithFullObject>(url, {
  headers: {
    'X-Get-Datasets-Full-Objects': 'True',
  },
  redirectOn404: true,
  redirectOnSlug: 'did',
})
```

nicolaskempf57:
> The dataset API has a special flag to return the full objects (and avoid doing too much work on the frontend. Have you try this?

I didn't try it and it worked ! So many avoided requests

PR #709, opened 2025-10-01
Add user activity

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/1876


agarrone:
@ThibaudDauce activity In my view, activities for a user or an org are not very critical. They are aimed at advanced users or admins. It’s fine if they’re not very prominent, unlike statistics, which must be easily accessible to the user, especially for organizations. And it makes sense, I think, that they should appear in the same way, in the same place, for both users and organizations.

PR #712, opened 2025-10-03
fix: custom header not sent

nicolaskempf57:
I don't have a clear understanding of why our custom header is not sent.
Maybe it's related to https://github.com/nuxt/nuxt/issues/33013 or to or usage of `await` in `useAPI` (not recommended).

I think this PR is good enough to allow us to deploy on dev / demo / prod our main branch and in the meantime we should investigate the replacement of our `await` with a promise based approach.


nicolaskempf57:
This implementation broke our `redirectOnSlug`, we'll revert the nuxt update instead.

PR #722, opened 2025-10-07
feat: add udata version in footer

ThibaudDauce:
Following https://github.com/opendatateam/udata/pull/3458


maudetes (approved):
Thank you!

PR #742, opened 2025-10-15
feat: don't display short description at all (for now) on dataset pages

bolinocroustibat:
After discussing with @agarrone and @maudetes we decided not to display short descriptions at all on dataset pages, since **for now** they are all cropped descriptions, which creates UI/wording duplicates.

We will change this policy later when/if short descriptions will differ more from descriptions.

Ideally this should be merged to main and to https://github.com/datagouv/cdata/pull/634.


maudetes (approved):
:pray:

nicolaskempf57 (approved):

PR #745, opened 2025-10-15
Stylize PMTiles preview popup

maudetes:
Clean the PMTiles preview popup.
See [the figma](https://www.figma.com/design/wFzskZXEyAhrt14Cdmydi3/Tooltip-pm-tiles?node-id=0-1&t=ePUNl9qYJodpPX0f-1).

### DIFF
--- a/datagouv-components/src/components/ResourceAccordion/Pmtiles.client.vue
+++ b/datagouv-components/src/components/ResourceAccordion/Pmtiles.client.vue
@@ -197,6 +197,9 @@ onMounted(() => {
 .maplibregl-popup-close-button {
   right: 10px !important;
   top: 10px !important;
-  font-size: 1.2rem;
+  width: 1rem;
+  height: 1rem;
+  font-size: 1rem;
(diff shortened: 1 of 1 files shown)


components/Datasets/DescribeDataset.vue
```diff
@@ -288,6 +294,55 @@ Elle aide les utilisateurs à comprendre rapidement ce qu’il contient et amél
...
+                </BrandedButton>
+                <template #tooltip>
+                  {{ $t('Remplissez le titre et une description d\'au moins {min} caractères pour utiliser cette fonctionnalité.', { min: DESCRIPTION_MIN_LENGTH }) }}
+                </template>
+              </Tooltip>
+              <BrandedButton
+                v-else
+                type="button"
+                color="primary"
+                :disabled="isGeneratingDescriptionShort"
```
nicolaskempf57: You can use the `loading` props instead of the `disabled` one for this, it will disable the button and add a spinner

PR #747, opened 2025-10-16
fix: don't throw error and change Albert LLM prompt if organization is not provided

bolinocroustibat:
Don't throw error and change Albert LLM prompt if organization is not provided.


maudetes (approved):
:ok_hand:

PR #749, opened 2025-10-17
chore: fix pre-commit issue, lint files and add lints to CI

nicolaskempf57:
Recent changes to cdata was likely done without the lint pre-commit.

This prevents contributor to commit to the project because the linter is complaining.

This PR tries to avoid this repo state by adding lint steps to the CI to avoid merging content not linted.

This is not perfect because the pre-commit use `lint-staged` with cdata config and the CI is using `lint` with the config of each project, with some differences between them.


maudetes (approved):
:muscle:

PR #779, opened 2025-11-24
feat(harvest): expose remote_url at the item level

maudetes:
Close https://github.com/datagouv/data.gouv.fr/issues/1763

Based on https://github.com/opendatateam/udata/pull/3538


streino (approved):
Approving as in "I think this is a great addition" :) I'll leave the actual code review to someone more familiar with cdata... :pray:

PR #781, opened 2025-11-24
feat: moderation admin page

ThibaudDauce:
- Require https://github.com/opendatateam/udata/pull/3539
- Fix https://github.com/datagouv/data.gouv.fr/issues/1904


pages/admin/site/moderation.vue
maudetes: At the moment I have multiple api calls by the client on the reported objects, ie `/api/2/datasets/61a770901d3bb2a227b5a6d3/`.

pages/admin/site/moderation.vue
```diff
@@ -0,0 +1,377 @@
...
+                    size="xs"
+                    color="tertiary"
+                    type="button"
+                    :icon="RiCheckLine"
+                    icon-only
+                    keep-margins-even-without-borders
+                    :disabled="!!report.dismissed_at"
+                    @click="dismiss(report)"
+                  >
+                    {{ $t('Rejeter le signalement') }}
```
maudetes: The labels don't show at the moment

pages/admin/site/moderation.vue
```diff
@@ -0,0 +1,377 @@
...
+
+const page = ref(1)
+const pageSize = ref(20)
+
+const filterStatus = ref({ label: t('En cours'), id: 'ongoing' })
+const filterStatusValue = computed(() => filterStatus.value.id || undefined)
+
+const { data: pageData, status, refresh } = await useAPI<PaginatedArray<Report>>(`/api/1/reports/`, { query: {
+  status: filterStatusValue,
+} })
```
maudetes: Maybe we should add a sort on report and show most recent reports first?

pages/admin/site/moderation.vue
```diff
@@ -0,0 +1,377 @@
...
+  }
+}, { immediate: true })
+
+const activities = ref({} as Record<string, Activity>)
+watchEffect(async () => {
+  if (!pageData.value) return
+
+  const subjects = pageData.value.data.map(r => r.subject as ReportSubject).filter(s => s)
+
+  const newActivities = await getLatestActivitiesForObjects($api, subjects)
```
maudetes: We're fetching the last activity but we're showing it in the column `Créé le`, which make it seems like it's the creation activity.

pages/admin/site/moderation.vue
```diff
@@ -0,0 +1,377 @@
...
+  await fetchFullSubject(report, subject)
+}
+
+const getDeleteUrl = (subject: ReportSubject): string | null => {
+  return {
+    Dataset: `/api/1/datasets/${subject.id}/`,
+    Dataservice: `/api/1/dataservices/${subject.id}/`,
+    Reuse: `/api/1/reuses/${subject.id}/`,
+    Organization: `/api/1/organizations/${subject.id}/`,
+    Discussion: null,
```
maudetes: Don't we have a delete url for discussions?

pages/admin/site/moderation.vue
ThibaudDauce: It's an `icon-only` button, the label is only here for accessibility reasons

pages/admin/site/moderation.vue
agarrone: There is 2 different date
- The objet creation date
- The trigger date

I'm not sure I understand

pages/admin/site/moderation.vue
maudetes: Could we have it on hover as well?

PR #805, opened 2025-12-04
fix: more TypeScript errors

ThibaudDauce:
Following https://github.com/datagouv/cdata/pull/787

Before: `Found 264 errors in 107 files.`. After: `Found 13 errors in 12 files.`

  - Separate API submission types from full entity types: When creating resources/harvesters, the API doesn't return server-generated fields (like permissions), so we need distinct types
  (NewResourceForApi, NewHarvesterForApi) for what we send vs what we receive
  - Support reference types alongside full types: Components receiving data from different sources (API responses, suggestions, forms) need to accept both OrganizationReference/UserReference
  and full Organization/User types
  - Handle enriched data in forms: Form fields can contain either raw API data or client-enriched versions (e.g., License vs EnrichedLicense with added group field for grouping in selects)
  - Improve type precision for callback props: Replace generic string parameters with literal unions so child components can use stricter function signatures that only accept their known keys
  - Fix avatar resolution for different user types: UserSuggest uses avatar_url while UserReference uses avatar_thumbnail, so the avatar helper needs to handle both patterns
  - Add missing required fields in resource forms: schema_url, checksum_type, checksum_value were missing from initial form state
  - Create centralized API fetch type: ApiFetch type replaces inconsistent use of $Fetch/$fetch from different packages
  - Switch pnpm from file: to link: protocol: Avoids reinstalling local package dependencies on every pnpm install
  - Add TypeScript assertions for Nuxt config options: Some valid Nuxt options (ssr in routeRules, api in scss) aren't in the types yet
  - Fix route params typing: Cast route.params to proper types where Nuxt infers string | string[]
  - Add null guards and non-null assertions: Protect against null values in places where TypeScript can't infer runtime guarantees

It's not the best solutions, but the goal is to be free of errors and we'll see to improve the hacks later.

### DIFF
--- a/components/Datasets/DescribeDataset.vue
+++ b/components/Datasets/DescribeDataset.vue
@@ -798,7 +798,7 @@ const { data: allLicenses } = await useAPI<Array<License>>('/api/1/datasets/lice
 
 // Merge some information between database (all licenses) and config (selectable license, some recommanded, codes…)
 // Maybe all these information could be better stored in database too…
-const licenses = computed((): Array<EnrichedLicense> => {
+const licenses = computed<Array<EnrichedLicense>>(() => {
   if (!allLicenses.value) return []
 
   const licenses: Array<EnrichedLicense> = []


components/Datasets/DatasetsMetrics.vue
nicolaskempf57: Il y a eu un problème de merge à un moment, la simplification des metrics a sauté (https://github.com/datagouv/cdata/pull/800/files#diff-768dc81905d93597554ae6137b21330126f6bda177e05af63141e45de83e24d6)

PR #812, opened 2025-12-09
feat: add onboarding pages

ThibaudDauce:
We can remove the pages / add the redirects later, so we can merge it fast and deploy it on dev/demo/production without exposing them.

---

image

---

image

---

image


frame.png
nicolaskempf57: to remove ?

nicolaskempf57 (commented):
I had to add `sharp` to handle png with nuxt-image (unrelated to this PR I guess). There are some unnecessary files to remove

PR #826, opened 2025-12-17
feat(datasets): add RSS feed for search results

bjperson:
## Summary

- Add RSS feed endpoint at `/datasets/search.xml`
- Supports all search filters (q, sort, tag, license, format, organization, owner, badge, geozone, granularity, topic)
- Default sort by creation date (newest first) when no sort specified
- Add RSS link button in the datasets search page header
- Link dynamically includes current search parameters

## Usage

```
/datasets/search.xml
/datasets/search.xml?q=transport
/datasets/search.xml?q=population&organization=534fff81a3a7292c64a77e5c
```

## Security

- XML escaping for user input in channel title
- Whitelisted parameters only (ALLOWED_PARAMS)
- page_size limited to 100 max
- CDATA sections for dataset content

## Test plan

- [ ] Access `/datasets/search.xml` without parameters (should show newest datasets)
- [ ] Access `/datasets/search.xml?q=transport` with search term
- [ ] Access with multiple filters (organization, tag, etc.)
- [ ] Click RSS button from search page, verify URL includes current filters
- [ ] Test with special characters in search (e.g., `<script>`)

### DIFF
--- a/components/Onboarding/OnboardingFeatureCard.vue
+++ b/components/Onboarding/OnboardingFeatureCard.vue
@@ -6,12 +6,12 @@
       class="rounded-t overflow-hidden flex items-center justify-center py-8 md:py-[60px] px-4 bg-gradient-to-b border-b border-gray-default"
       :class="gradientClass"
     >
-      <NuxtImg
+      <img
         :src="image"
         alt=""
         class="h-[240px] md:h-[320px] w-auto object-contain"
         loading="lazy"
-      />
+      >
     </div>
     <!-- Petite carte : container avec aspect-ratio, image proportionnelle -->
     <div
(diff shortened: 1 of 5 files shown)


bjperson:
Hi, thanks for reviewing!

My use case is **targeted monitoring of new datasets** directly from an RSS reader (Feedly, Inoreader, Thunderbird...) without having to build an API client.

The existing "recent datasets" feed shows all new datasets globally, but doesn't allow filtering. With this PR, users can subscribe to specific searches like:
- `/datasets/search.xml?q=transport` - monitor "transport" datasets
- `/datasets/search.xml?organization=534fff81a3a7292c64a77e5c` - follow a specific organization
- `/datasets/search.xml?tag=environnement&format=csv` - combine multiple filters

This enables data journalists, researchers, and citizens to set up automated alerts on topics they care about, using standard RSS tools they already have.

The JSON API is great for developers, but RSS lowers the barrier for non-technical users who just want to stay informed.

maudetes:
I think we should implement filtering on the `api/1/datasets/recent.atom` endpoint directly instead.
I guess it would be in phase with implementation on rdf catalog added in https://github.com/opendatateam/udata/pull/3474 as well.

ThibaudDauce:
Thanks! The udata project is here https://github.com/opendatateam/udata/

nicolaskempf57:
> I'm happy to close this PR and contribute to udata instead if you can point me to the relevant code for the Atom feed implementation.

The Atom feed is implemented here : https://github.com/opendatateam/udata/blob/main/udata/core/dataset/api.py#L328

PR #845, opened 2025-12-23
feat: add markdown bloc in pages

ThibaudDauce:
- Require https://github.com/opendatateam/udata/pull/3581


nicolaskempf57 (approved):
We'll have a pass to do on all blocs to fix heading levels but it should be for all blocs not only this one

PR #854, opened 2026-01-06
chore: add test for upload chunks

ThibaudDauce:
Following https://github.com/opendatateam/udata/pull/3597 then the replacement by https://github.com/opendatateam/udata/pull/3598 (waiting werkzeug fix), we don't need to change the code but adding a test.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           repository: opendatateam/udata
           path: udata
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         with:
           repository: opendatateam/udata
           path: udata
-          ref: main
+          ref: upgrade_dependencies_and_downgrade_werkzeug
```
nicolaskempf57:
```suggestion
          ref: main
```

PR #861, opened 2026-01-08
feat: wysiwyg blocs editor

ThibaudDauce:
To add the accordion bloc, since there is embed blocs, it's too hard to do with the existing system (modal inside modal inside modal…), so I think it's time to migrate to a real Wysiwyg system where it will be easier (maybe) to do the edition of accordions :-)

image

### DIFF
--- a/components/Pages/BrandedButtonEditor.vue
+++ b/components/Pages/BrandedButtonEditor.vue
@@ -14,20 +14,18 @@
     >
       <RiDeleteBinLine class="size-5" />
     </button>
-    <BrandedButton
-      :color
-      class="pointer-events-none"
+    <div
+      class="inline-flex items-center justify-center font-marianne font-medium border text-sm h-10 leading-none px-4 gap-x-1"
+      :class="buttonColors"
     >
-      <input
-        :value="title"
-        type="text"
-        class="bg-transparent border-none outline-none text-inherit font-inherit pointer-events-auto"
-        :style="{ width: `${Math.max(10, title?.length || 0)}ch` }"
-        :placeholder="$t('Titre du bouton')"
+      <span
+        contenteditable="true"
+        role="textbox"
+        class="outline-none min-w-[10ch] focus:ring-2 focus:ring-blue-300 rounded-sm"
         :aria-label="$t('Titre du bouton')"
-        @input="$emit('update:title', ($event.target as HTMLInputElement).value)"
-      >
-    </BrandedButton>
+        @blur="$emit('update:title', ($event.target as HTMLElement).textContent || '')"
+      >{{ title || $t('Titre du bouton') }}</span>
+    </div>
     <input
       :value="href"
       type="text"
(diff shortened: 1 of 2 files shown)


components/Pages/BrandedButtonEditor.vue
```diff
@@ -0,0 +1,71 @@
...
+      :title="$t('Supprimer')"
+      @click="$emit('clear')"
+    >
+      <RiDeleteBinLine class="size-5" />
+    </button>
+    <BrandedButton
+      :color
+      class="pointer-events-none"
+    >
+      <input
```
nicolaskempf57: Our inputs should have `aria-label` at least if they don't have proprer `<label>`

components/Pages/EditableText.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <component
+    :is="tag"
+    :contenteditable="true"
```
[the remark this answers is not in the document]
ThibaudDauce: For now it's super admin only so it's not really a concern I think, we could add an accessibility mode with real inputs easily in the futur I think (or even for everyone but right now I couldn't make it work)

components/Pages/PageShow.vue
```diff
@@ -1,117 +1,233 @@
...
+        class="absolute -top-4 left-0 right-0 flex items-center justify-center z-20"
       >
-        <DatasetCardLg
-          v-for="dataset in bloc.datasets"
-          :key="dataset.id"
-          class="min-w-0"
-          :dataset
-        />
+        <AddBlocDropdown @new-bloc="workingPage.blocs.splice(index, 0, $event)">
+          <BrandedButton
```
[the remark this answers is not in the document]
ThibaudDauce: Fix in 96abad0

components/Pages/BrandedButtonEditor.vue
```diff
@@ -0,0 +1,73 @@
...
+    <button
+      v-if="title || href"
+      type="button"
+      class="absolute -left-12 top-1/2 -translate-y-1/2 text-gray-500 hover:text-gray-700"
+      :title="$t('Supprimer')"
+      @click="$emit('clear')"
+    >
+      <RiDeleteBinLine class="size-5" />
+    </button>
+    <BrandedButton
```
nicolaskempf57: I think we can use a contenteditable here with default value instead of the input. Otherwise we have to deal with the fact that the button has no text content.

ThibaudDauce: Done in 9237304 not sure if it's better but it works :-) Could improve all these contenteditable in a next PR…

nicolaskempf57: Couldn't you keep the Brandedbutton ? The idea was that it doesn't require many changes

ThibaudDauce: contenteditable in branded button is not much better I think? And without the branded button we can have a more simple HTML (without the hover states, etc.)

PR #873, opened 2026-01-15
feat: add elections page

ThibaudDauce:
image

Capture d’écran 2026-01-15 à 14 28 01

### DIFF
--- a/pages/elections.vue
+++ b/pages/elections.vue
@@ -50,6 +50,8 @@
 </template>
 
 <script setup lang="ts">
+import ElectionCard from '~/components/Elections/ElectionCard.vue'
+
 const config = useRuntimeConfig()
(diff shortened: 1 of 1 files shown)


components/Elections/ElectionCard.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <NuxtLink
+    :to="to"
+    class="flex flex-col md:flex-row gap-2.5 p-4 rounded-sm border border-gray-silver hover:border-gray-medium transition-colors bg-white bg-none"
+  >
+    <img
+      :src="image"
+      :alt="title"
```
nicolaskempf57:
```suggestion
      alt=""
```

I think alt should be empty here because the title is already shown below

PR #847, opened 2025-12-25
feat: refacto search with facets and multiple configs

geoffreyaldebert:

PR #913, opened 2026-02-03
feat: add new filters and facets to global search

ThibaudDauce:
Related to https://github.com/datagouv/data.gouv.fr/issues/1432

Follow up of https://github.com/datagouv/cdata/pull/907

Replace https://github.com/datagouv/cdata/pull/847

### DIFF
--- a/components/User/TwoFactorSetupModal.vue
+++ b/components/User/TwoFactorSetupModal.vue
@@ -111,15 +111,12 @@ const {
 } = useTwoFactorSetup()
 
 const fetchStatusAndQRCode = async () => {
-  const { data: twoFactorData } = await useAPI<{ response: { tf_primary_method: string | null, reauth_required: boolean } | null }>('/tf-setup')
-  isConfigured.value = !!twoFactorData.value?.response?.tf_primary_method
-  if (twoFactorData.value?.response?.reauth_required) {
-    toast.error(t('Vous devez vous déconnecter et reconnecter pour pouvoir configurer l\'authentification deux facteurs.'))
-    isOpen.value = false
-    return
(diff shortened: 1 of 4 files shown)


datagouv-components/src/components/Search/Filter/DatasetBadgeFilter.vue
```diff
@@ -0,0 +1,37 @@
...
+  loading?: boolean
+}>()
+
+const emit = defineEmits<{
+  'update:modelValue': [value: string | undefined]
+}>()
+
+const { t } = useTranslation()
+
+const options = [
```
nicolaskempf57: I think we should use `api/1/datasets/badges/` to get the labels and the existing `datasetBadges` config to filter the ones we want to show, like the previous dataset search

ThibaudDauce: Use API in cb933368c6a458528b6d7b8539ecace7d86912f0, for the filter I don't like the idea of having a frontend config for that and I think we should find a way to filter on udata. I don't think it's a problem for this PR to show "Données pivots" with 0 results for now?

datagouv-components/src/types/search.ts
```diff
@@ -350,7 +351,8 @@ export function getDefaultDatasetConfig(overrides?: Partial<Omit<DatasetSearchCo
 export function getDefaultDataserviceConfig(overrides?: Partial<Omit<DataserviceSearchConfig, 'class'>>): DataserviceSearchConfig {
   return {
     class: 'dataservices',
-    basicFilters: ['organization', 'is_restricted'],
+    basicFilters: ['access_type', 'last_update_range', 'producer_type'],
```
maudetes: I think `last_update_range` does not work on API, can we remove it for now?

PR #920, opened 2026-02-05
feat(2fa): add security verify route

maudetes:
Built on https://github.com/datagouv/cdata/pull/921

Require https://github.com/opendatateam/udata/pull/3644

### DIFF
--- a/tests/auth.2fa-user.spec.ts
+++ b/tests/auth.2fa-user.spec.ts
@@ -92,7 +92,7 @@ test.describe('2FA Authentication Flow', () => {
 
     // Step 12: Verify error message is displayed
     await expect(page.locator('.fr-error-text')).toBeVisible()
-    await expect(page.getByText('Code invalide')).toBeVisible()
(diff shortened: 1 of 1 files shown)


plugins/api.ts
```diff
@@ -45,7 +45,12 @@ export default defineNuxtPlugin({
           }
 
           if (response.status === 401) {
-            await nuxtApp.runWithContext(() => navigateTo({ path: '/login', query: { next: route.fullPath } }))
+            if (response._data?.response && typeof response._data.response === 'object' && 'reauth_required' in response._data.response) {
+              await nuxtApp.runWithContext(() => navigateTo({ path: '/verify', query: { next: route.fullPath } }))
+            }
+            else {
+              await nuxtApp.runWithContext(() => navigateTo({ path: '/login', query: { next: route.fullPath } }))
+            }
```
ThibaudDauce: Maybe add a return after this line, I'm not sure what the await / navigateTo do after the promise…

```suggestion
            }
            return
```

PR #963, opened 2026-02-26
feat: add viz poc 1

### LINKED ISSUE #1972: 📊 Datastudio sur data.gouv.fr
Design : [page Flow](https://www.figma.com/design/423gxuZAVkoGJ5ShIyPRV2/%F0%9F%9B%B8-DATASTUDIO-DU-FUTUR?node-id=193-3530&p=f&t=HNfyrdKxyagC1h01-0)
Proto : https://datagouvdatastudio.figma.site/

nicolaskempf57:
Fixes https://github.com/datagouv/data.gouv.fr/issues/1972


package.json
```diff
@@ -90,6 +90,7 @@
     "uqr": "^0.1.2",
     "uuid": "^13.0.0",
     "vue-content-loader": "^2.0.1",
+    "vue-echart": "^1.0.5",
```
ThibaudDauce: This is the wrong package (singuler instead of plural), getting Vue 2 and an old axios… Never used

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  const getPromise = page.waitForResponse(response => response.url().includes(`${getApiBase()}/api/1/visualizations/`) && response.request().method() === 'GET')
+
+  await page.getByRole('button', { name: 'Sauvegarder le graphique' }).click()
+  const response = await responsePromise
+  const responseBody = (await response.json()) as Chart
+  await getPromise
+
+  expect(responseBody!.title).toBe('Test Chart')
+  expect(await page.getByLabel('Graphiques existants').inputValue()).toBe(responseBody!.id)
+  await page.request.delete(`${getApiBase()}/api/1/visualization/${responseBody!.id}/`)
```
ThibaudDauce:
```suggestion
  await page.request.delete(`${getApiBase()}/api/1/visualizations/${responseBody!.id}/`)
```

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  expect(await page.getByLabel('Type', { exact: true }).first().inputValue()).toBe('continuous')
+  expect(await page.getByLabel('Colonne Y').inputValue()).toBe('Parc social - Taux de logements vacants* (en %)')
+  expect(await page.getByLabel('Agrégation').inputValue()).toBe('avg')
+  expect(await page.getByLabel('Label').inputValue()).toBe('Taux (%)')
+  expect(await page.getByLabel('Min').inputValue()).toBe('0')
+  expect(await page.getByLabel('Max').inputValue()).toBe('100')
+  expect(await page.getByLabel('Unité', { exact: true }).inputValue()).toBe('%')
+  expect(await page.getByLabel('Position unité').inputValue()).toBe('prefix')
+
+  await page.request.delete(`${getApiBase()}/api/1/visualization/${responseBody!.id}/`)
```
ThibaudDauce:
```suggestion
  await page.request.delete(`${getApiBase()}/api/1/visualizations/${responseBody!.id}/`)
```

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+import { getApiBase } from '../helpers'
+
+test('dataset selector shows default dataset', async ({ page }) => {
+  await page.goto('/design/chart')
+  await page.waitForLoadState('networkidle')
+
+  const datasetSelect = page.getByPlaceholder('Recherchez un jeu de données...')
+  // SearchableSelect displays the selected value in a different way
+  const displayedValue = await datasetSelect.inputValue()
+  expect(datasetSelect).toBeVisible()
```
ThibaudDauce:
```suggestion
  await expect(datasetSelect).toBeVisible()
```

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  expect(displayedValue || datasetSelect.inputValue()).toContain('Logements et logements sociaux')
```
ThibaudDauce: Should also await datasetSelect.inputValue ?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+  return {
+    '': t('Aucun'),
+    'axis_x-asc': xAxisColumn ? `${xAxisColumn} - ${t('Ascendant')}` : t('Axe X - Ascendant'),
+    'axis_x-desc': xAxisColumn ? `${xAxisColumn} - ${t('Descendant')}` : t('Axe X - Descendant'),
+    'axis_y-asc': yAxisColumn ? `${yAxisColumn} - ${t('Ascendant')}` : t('Axe Y - Ascendant'),
+    'axis_y-desc': yAxisColumn ? `${yAxisColumn} - ${t('Descendant')}` : t('Axe Y - Descendant'),
+  }
+})
+
+const columnDetails = computed(() => [{ key: '', value: t('Colonne'), disabled: true }, ...columns.value[selectedResource.value].map(c => ({ key: c, value: c, disabled: false }))])
```
ThibaudDauce: `columns.value[selectedResource.value]` could be null before loading?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+
+function getColumnOptions() {
+  return columnDetails.value
+}
```
ThibaudDauce: What the point of this?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+}, { immediate: true })
+
+watch([titleDebounced, descDebounced], ([title, desc]) => {
+  form.value.title = title
+  form.value.description = desc
+})
+
+watch(columns, (columnsPerResource) => {
+  const firstColumns = Object.values(columnsPerResource) ?? []
+  const firstColumn = firstColumns[0].filter(c => c !== '__id')[0]
```
ThibaudDauce, line 503: Could `firstColumns` be empty in some cases?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+  }
+}, { immediate: true })
+
+watch([titleDebounced, descDebounced], ([title, desc]) => {
+  form.value.title = title
+  form.value.description = desc
+})
+
+watch(columns, (columnsPerResource) => {
+  const firstColumns = Object.values(columnsPerResource) ?? []
```
ThibaudDauce: Why `?? []` is useful here?

nuxt.config.ts
```diff
@@ -69,7 +69,7 @@ export default defineNuxtConfig({
       maxPdfPreviewByteSize: 10000000, // (10 MB)
       maxXmlPreviewCharSize: 100000, // (~100KB)
       schemaValidataUrl: 'https://validata.fr',
-      tabularApiUrl: 'https://tabular-api.data.gouv.fr',
+      tabularApiUrl: 'https://tabular-api.preprod.data.gouv.fr',
```
ThibaudDauce: I think we should revert this? You can use your `.env` in testing?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+const hasTabularData = useHasTabularData()
+const getProfile = useGetProfile()
+const isAdmin = isMeAdmin()
+
+const { data: charts, refresh: refresh } = await useAPI<PaginatedArray<Chart>>('/api/1/visualizations/', { lazy: true })
+
+watch(dataset, async (newDataset) => {
+  if (!newDataset) return
+  try {
+    const fetchedResources = await $api<PaginatedArray<Resource>>(`https://demo.data.gouv.fr/api/2/datasets/${newDataset.id}/resources/`)
```
ThibaudDauce: "https://demo.data.gouv.fr"?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+    form.value.x_axis.column_x = firstColumn
+  }
+})
+
+async function loadMissingResourcesForChart(resourceIds: Set<string>) {
+  for (const id of resourceIds) {
+    if (savedResources[id]) continue
+
+    try {
+      const resource = await $api<{ resource: Resource, dataset_id: string }>(`https://demo.data.gouv.fr/api/2/datasets/resources/${id}/`)
```
ThibaudDauce: "https://demo.data.gouv.fr" to remove

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,595 @@
...
+      savedResources[id] = resource.resource
+    }
+    catch (error) {
+      console.error(`Failed to fetch resource ${id}:`, error)
+    }
+  }
+}
+
+async function suggestDataset(q: string): Promise<Array<DatasetSuggest>> {
+  return await $api<Array<DatasetSuggest>>('https://demo.data.gouv.fr/api/1/datasets/suggest/', {
```
ThibaudDauce: "https://demo.data.gouv.fr" to remove

datagouv-components/src/functions/tabularApi.ts
```diff
@@ -1,26 +1,152 @@
...
+ * Call Tabular-api to get table content with options object
  */
-export async function getData(config: PluginConfig, id: string, page: number, sortConfig?: SortConfig) {
-  let url = `${config.tabularApiUrl}/api/resources/${id}/data/?page=${page}&page_size=${config.tabularApiPageSize || 15}`
-  if (sortConfig) {
-    url = url + `&${sortConfig.column}__sort=${sortConfig.type}`
+export async function fetchTabularData(config: PluginConfig, options: FetchTabularDataOptions): Promise<TabularDataResponse> {
+  const page = options.page ?? 1
+  const pageSize = options.pageSize ?? config.tabularApiPageSize ?? 15
+  console.log(options)
```
ThibaudDauce:
```suggestion
```

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,224 @@
...
+    series.columns = Object.fromEntries(results.map(result => [
+      result.id,
+      result.profile.profile.header,
+    ]))
+    status.value = 'success'
+  }
+  catch (err) {
+    error.value = err instanceof Error ? err : new Error('Failed to fetch series profile')
+    status.value = 'error'
+    console.log(err)
```
ThibaudDauce:
```suggestion
    console.error(err)
```

components/Datasets/FileEditModal.vue
```diff
@@ -228,6 +228,7 @@ const resourceForm = ref(cloneDeep(props.resource))
 const open = ref(false)
 const hasFileChanged = ref(false)
 
+// Use the composable for tabular data detection
```
ThibaudDauce: Is this comment useful?

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,147 @@
...
+
+    if (!xColumn || !yColumn || !resourceId || !s.type || !props.series.data[resourceId] || !props.series.columns[resourceId]) {
+      return null
+    }
+
+    const sortedData = [...props.series.data[resourceId]]
+    let sortBy: string | null = null
+    let sortDirection: 'asc' | 'desc' = 'asc'
+    sortBy = props.chart.x_axis.sort_x_by
+    sortDirection = props.chart.x_axis.sort_x_direction ?? 'asc'
```
ThibaudDauce: Why separate init from setting the value?

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  await expect(saveButton).toBeEnabled()
+})
+
+test('saving chart sends correct data to API', async ({ page }) => {
+  await page.goto('/design/chart')
+  await page.waitForLoadState('networkidle')
+
+  await page.getByLabel('Titre').fill('Test Chart')
+  await page.getByLabel('Description').fill('Test Description')
+  await page.waitForTimeout(500)
```
ThibaudDauce, line 100: Do we need this timeout? Could we have something more precise?

tests/visualizations/chart-configurator.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+})
+
+test('save button is visible and clickable', async ({ page }) => {
+  await page.goto('/design/chart')
+  await page.waitForLoadState('networkidle')
+
+  const saveButton = page.getByRole('button', { name: 'Sauvegarder' })
+  await expect(saveButton).toBeVisible()
+  await expect(saveButton).toBeEnabled()
+})
```
ThibaudDauce: Is this test really useful?

datagouv-components/src/composables/useHasTabularData.ts
```diff
@@ -1,15 +1,23 @@
...
+  const hasTabularData = (resource: Resource) => {
     return (
       config.tabularApiUrl
       && resource.extras['analysis:parsing:parsing_table']
       && !resource.extras['analysis:parsing:error']
       && (config.tabularAllowRemote || resource.filetype === 'file')
     )
   }
+
+  return hasTabularData
```
ThibaudDauce: To revert?

datagouv-components/src/components/ResourceAccordion/Preview.vue
```diff
@@ -142,7 +142,7 @@ async function getTableInfos(page: number, sortConfig?: SortConfig) {
   try {
     // Check that this function return wanted data
     const response = await getData(config, props.resource.id, page, sortConfig)
-    if ('data' in response && response.data && response.data.length > 0) {
+    if ('data' in response && response.data && 0 in response.data) {
```
ThibaudDauce, line 145: Why this change?

ThibaudDauce (changes requested):

tests/visualizations/chart-configurator.spec.ts
nicolaskempf57: I'll simplify the test, I think I changed the start but not the end

components/Charts/ChartConfigurator.vue
nicolaskempf57: A leftover of a previous implementation

tests/visualizations/chart-configurator.spec.ts
nicolaskempf57: There is a debounce for the title / description

datagouv-components/src/components/ResourceAccordion/Preview.vue
nicolaskempf57: I think it was a type issue, but I will double check

tests/visualizations/chart-configurator.spec.ts
nicolaskempf57: I changed it to 300, I don't think we can do it differently

datagouv-components/src/components/Chart/ChartViewer.vue
nicolaskempf57: I think it's a leftover of a previous attempt

nuxt.config.ts
nicolaskempf57: Yep

components/Charts/ChartConfigurator.vue
nicolaskempf57: Removed, columns is always an object

components/Charts/ChartConfigurator.vue
nicolaskempf57: Technically it could, condition added

tests/visualizations/chart-configurator.spec.ts
ThibaudDauce: There is no text appearing after the debounce?

nicolaskempf57: The chart title is updated but we can't select it for now, it's inside the canvas. Later it'll be replace with a p tag and one for the description too but this is not done yet

datagouv-components/src/components/ResourceAccordion/Preview.vue
nicolaskempf57: Yep, the issue from the typecheck is :

```
Error: src/components/ResourceAccordion/Preview.vue(148,35): error TS2769: No overload matches this call.
  Overload 1 of 2, '(o: {}): string[]', gave the following error.
    Argument of type 'Record<string, unknown> | undefined' is not assignable to parameter of type '{}'.
      Type 'undefined' is not assignable to type '{}'.
  Overload 2 of 2, '(o: object): string[]', gave the following error.
    Argument of type 'Record<string, unknown> | undefined' is not assignable to parameter of type 'object'.
      Type 'undefined' is not assignable to type 'object'.
```

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    // Fetch data for all series in parallel
+    const fetchPromises = props.chart.series.map(async (serie) => {
+      if (!serie.resource_id) return
+      return {
+        id: serie.resource_id,
+        profile: await getProfile(serie.resource_id),
+      }
+    }).filter(Boolean) as Array<Promise<{
```
ThibaudDauce: The .filter(Boolean) filter the `Promise` object so it's always `true`

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,224 @@
...
+          aggregation: serie.column_y && serie.aggregate_y
+            ? {
+                column: serie.column_y,
+                type: serie.aggregate_y,
+              }
+            : undefined,
+          filters: serie.filters ?? undefined,
+        }),
+      }
+    }).filter(Boolean) as Array<Promise<{
```
ThibaudDauce: Same here, `.filter(Boolean)` is called on the `Promise` so it's always `true`

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,145 @@
...
+      text: props.chart.title,
+      left: 'center',
+    },
+    tooltip: {
+      trigger: 'axis' as const,
+      formatter: (params: Array<{ value: Record<string, unknown>, axisValueLabel: string, seriesName: string }>) => {
+        let tooltip = ''
+        for (const param of params) {
+          const keys = Object.keys(param.value)
+          const col = keys.find(key => key.startsWith(param.seriesName))!
```
ThibaudDauce: `startsWith` is not the correct way to do that, no? If two columns starts with the same prefix like `pop` and `population` or if we have `population__avg` and `population__sum` in the same viz?

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,145 @@
...
+
+    const sortedData = [...props.series.data[resourceId]]
+    const sortBy = props.chart.x_axis.sort_x_by
+    const sortDirection = props.chart.x_axis.sort_x_direction ?? 'asc'
+
+    if (sortBy && sortDirection && props.chart.x_axis.column_x) {
+      const sortKey = sortBy === 'axis_x' ? xColumn : yColumn
+      sortedData.sort((a, b) => {
+        const valA = a[sortKey] as number
+        const valB = b[sortKey] as number
```
ThibaudDauce: There is a way to have something else than a number here? I think with discrete in X axis? Or with numbers as string `"10" < "2"` is wrong, no?

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,224 @@
...
+
+// Watch for changes in the chart or its series
+watch(() => props.chart.series, async () => {
+  await fetchSeriesProfile()
+}, { immediate: true, deep: true })
+
+// Watch for changes in the chart or its series
+watch([() => props.chart.series, () => props.chart.x_axis.column_x], async () => {
+  await fetchSeriesData()
+}, { immediate: true, deep: true })
```
ThibaudDauce, line 157: `fetchSeriesProfile` and `fetchSeriesData` run in parallel I think and they both reset and write in `status.value` and `error.value`, one error followed by one success will override the status?

datagouv-components/src/functions/tabularApi.ts
```diff
@@ -1,24 +1,153 @@
...
-export async function getData(config: PluginConfig, id: string, page: number, sortConfig?: SortConfig | null) {
-  let url = `${config.tabularApiUrl}/api/resources/${id}/data/?page=${page}&page_size=${config.tabularApiPageSize || 15}`
-  if (sortConfig) {
-    url = url + `&${sortConfig.column}__sort=${sortConfig.direction}`
+export async function fetchTabularData(config: PluginConfig, options: FetchTabularDataOptions): Promise<TabularDataResponse> {
+  const page = options.page ?? 1
+  const pageSize = options.pageSize ?? config.tabularApiPageSize ?? 15
+  let url = `${config.tabularApiUrl}/api/resources/${options.resourceId}/data/?page=${page}&page_size=${pageSize}`
+  if (options.columns) {
+    url += `&columns=${options.columns.join(',')}`
```
ThibaudDauce: Should we encode parameters here?

datagouv-components/src/functions/tabularApi.ts
```diff
@@ -1,24 +1,153 @@
...
+  }
+  if (options.sort) {
+    url += `&${options.sort.column}__sort=${options.sort.direction}`
```
ThibaudDauce: Same here?

datagouv-components/src/functions/tabularApi.ts
```diff
@@ -1,24 +1,153 @@
...
   }
-  return await ofetch(url)
+  if (options.groupBy && options.aggregation?.type) {
+    url += `&${options.groupBy}__groupby&${options.aggregation.column}__${options.aggregation.type}`
```
ThibaudDauce: Same here?

datagouv-components/src/functions/tabularApi.ts
```diff
@@ -1,24 +1,153 @@
...
+  }
+  if (options.filters) {
+    const filterQuery = buildFilterQuery(options.filters)
+    if (filterQuery) {
+      url += `&${filterQuery}`
```
ThibaudDauce, line 95: Same here?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,612 @@
...
+        </div>
+      </fieldset>
+
+      <div
+        v-if="isAdmin"
+        class="px-6"
+      >
+        <BrandedButton @click="saveChart">
+          {{ $t('Sauvegarder le graphique') }}
+        </BrandedButton>
```
ThibaudDauce: Should it be a `@submit` with a form?

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,224 @@
...
+  >
+    <template #default="{ data }">
+      <ChartViewer
+        :chart="chart"
+        :series="data"
+      />
+    </template>
+    <template #error>
+      <div class="text-center py-8 text-gray-500">
+        Une erreur est survenue lors du chargement des données du graphique.
```
ThibaudDauce:
```suggestion
        $t('Une erreur est survenue lors du chargement des données du graphique.')
```

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,612 @@
...
+        <SearchableSelect
+          v-model="dataset"
+          :label="$t('Jeu de données')"
+          :placeholder="$t('Recherchez un jeu de données...')"
+          :display-value="(option: DatasetSuggest) => option.title"
+          :get-option-id="(option: DatasetSuggest) => option.id"
+          :multiple="false"
+          :suggest="suggestDataset"
+          class="mb-4"
+        />
```
ThibaudDauce: Could we use `<DatasetSelect :single />`?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,612 @@
...
+const ChartViewerWrapper = defineAsyncComponent(() => import('@datagouv/components-next').then(m => m.ChartViewerWrapper))
+
+const form = defineModel<ChartForm>({
+  required: true,
+})
+
+const { t } = useTranslation()
+
+const columns = ref<Record<string, Array<string>>>({})
+const flattenedColumns = computed(() => Object.entries(columns.value).map(([key, value]) => ({ key, value })).flat())
```
ThibaudDauce: Is the `.flat()` useful?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,612 @@
...
+const $chartsApi = $fetch.create({
+  baseURL: runtimeConfig.public.chartsApiBase,
+  onRequest({ options }) {
+    options.headers.set('Content-Type', 'application/json')
+    options.headers.set('Accept', 'application/json')
+    options.credentials = 'include'
+  },
+})
+
+const { data: charts, refresh: refresh } = await useAPI<PaginatedArray<Chart>>('/api/1/visualizations/', { lazy: true })
```
ThibaudDauce:
```suggestion
const { data: charts, refresh } = await useAPI<PaginatedArray<Chart>>('/api/1/visualizations/', { lazy: true })
```

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,145 @@
...
+      left: 'center',
+    },
+    tooltip: {
+      trigger: 'axis' as const,
+      formatter: (params: Array<{ value: Record<string, unknown>, axisValueLabel: string, seriesName: string }>) => {
+        let tooltip = ''
+        for (const param of params) {
+          const keys = Object.keys(param.value)
+          const col = keys.find(key => key.startsWith(param.seriesName))!
+          const formatter = new Intl.NumberFormat('fr-FR')
```
ThibaudDauce: Could be outside the loop

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,145 @@
...
+          x: xColumn,
+          y: yColumn,
+        },
+      } as LineSeriesOption | BarSeriesOption,
+      data: {
+        source: sortedData,
+        dimensions: s.aggregate_y ? [xColumn, yColumn] : props.series.columns[resourceId],
+      },
+    }
+  }).filter(Boolean).reduce((acc: { series: Array<LineSeriesOption | BarSeriesOption>, data: Array<Record<string, unknown>> }, curr) => {
```
ThibaudDauce: The `.reduce` here is complex, maybe just do a simple loop?

ThibaudDauce (changes requested):

components/Charts/ChartConfigurator.vue
nicolaskempf57: Not really, we don't want a complex component with card preview, just the `SearchableSelect`

nicolaskempf57: I tried it, but because of the design in the sidebar, it was strange

components/Charts/ChartConfigurator.vue
nicolaskempf57: Done

components/Charts/ChartConfigurator.vue
nicolaskempf57: No, I think it was a leftover of a previous implementation

datagouv-components/src/components/Chart/ChartViewer.vue
nicolaskempf57: Yes, i added more sort cases

datagouv-components/src/components/Chart/ChartViewer.vue
nicolaskempf57: Done !

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
nicolaskempf57: I added a simple count to have a success only when all succeed

datagouv-components/src/components/Chart/ChartViewer.vue
nicolaskempf57: Done by computing the aggregation column name

Velawan2823 (approved):

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,619 @@
...
+                {{ column.title }}
+              </option>
+            </select>
+            <button
+              class="fr-btn"
+              :disabled="!selectedChartId"
+              @click="loadSelectedChart"
+            >
+              {{ $t('Charger') }}
+            </button>
```
ThibaudDauce, line 82: Missing `type="button"`, no?

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,226 @@
...
+      left: 'center',
+    },
+    tooltip: {
+      trigger: 'axis' as const,
+      formatter: (params: Array<{ value: Record<string, unknown>, axisValueLabel: string, seriesName: string }>) => {
+        let tooltip = ''
+        const formatter = new Intl.NumberFormat('fr-FR')
+        for (const param of params) {
+          const seriesName = param.seriesName
+          const seriesConfig = props.chart.series.find(s => s.column_y === seriesName)
```
ThibaudDauce: If there is two series with the same column name (two different aggregate sum/mean for exemple), you'll always get the first one and then `${seriesName}__${seriesConfig.aggregate_y}` could match nothing? (the seriesName from the one you hover but the `seriesConfig.aggregate_y` from the first serie?

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
ThibaudDauce: I'm not sure it solves all the problems with this design. Setting `status = 'success'` after fetching the profile will show a flash of the chart without dataset then `fetchSeriesData` resolve and the chart is populated? Maybe remove `status = 'success'` and let the `watch(pendingOperation)` set it to be sure the two fetches are done before setting the status or maybe refactor this logic to be simpler? `fetchSeriesProfile` and `fetchSeriesData` are only called in these watchers? You may merge them together and call with a `Promise.all()`. If you don't want to get the profile on `props.chart.x_axis.column_x` you may just have one function with an argument `dataOnly=true`, it will solves all these problems, no?

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,619 @@
...
+}
+
+async function suggestDataset(q: string): Promise<Array<DatasetSuggest>> {
+  return await $api<Array<DatasetSuggest>>('/api/1/datasets/suggest/', {
+    query: { q, size: 5 },
+  })
+}
+
+function removeFilter(index: number) {
+  if (!form.value.filter || !('filters' in form.value.filter)) return
```
ThibaudDauce: `removeFilter` returns early when `form.value.filter` is a bare `Filter`
(not an `AndFilters`), but `filterList` (line 597) still renders a row
with its delete button in that case — clicking it is a silent no-op.

The form lands in this state when loading an existing chart whose
`series[0].filters` is a bare `Filter`: `toChartForm` (charts.ts:4)
casts it `as Filter | null` and assigns it directly to
`form.value.filter`, even though `DataSeries.filters` is typed as
`GenericFilter` and can be either shape at runtime.

Two options:
- normalize in `toChartForm` so `form.value.filter` is always either
  `null` or an `AndFilters` (then `filterList` and `removeFilter` lose
  the dual-shape branching),
- or handle the bare-Filter case here:

  if (!form.value.filter) return
  if (!('filters' in form.value.filter)) {
    if (index === 0) form.value.filter = null
    return
  }
  form.value.filter.filters.splice(index, 1)
  ...

The normalization in `toChartForm` is cleaner.

datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -0,0 +1,226 @@
...
+    if (sortBy && sortDirection && props.chart.x_axis.column_x) {
+      const sortKey = sortBy === 'axis_x' ? xColumn : yColumn
+      sortedData.sort((a, b) => {
+        const valA = a[sortKey]
+        const valB = b[sortKey]
+
+        if (valA === null || valA === undefined) return sortDirection === 'asc' ? -1 : 1
+        if (valB === null || valB === undefined) return sortDirection === 'asc' ? 1 : -1
+        if (valA === null && valB === null) return 0
+        if (valA === undefined && valB === undefined) return 0
```
ThibaudDauce: The "both null" / "both undefined" checks on lines 72-73 are unreachable:
lines 70-71 already return as soon as either side is null/undefined, so
when both values are null the comparator returns -1 (asc) instead of 0 —
a violation of comparator equality. Modern engines use a stable sort so
the practical impact is limited, but the dead checks reveal the intent
doesn't match the behavior.

Either drop lines 72-73, or (cleaner) handle equality first:

  const aNullish = valA === null || valA === undefined
  const bNullish = valB === null || valB === undefined
  if (aNullish && bNullish) return 0
  if (aNullish) return sortDirection === 'asc' ? -1 : 1
  if (bNullish) return sortDirection === 'asc' ? 1 : -1

datagouv-components/src/components/Chart/ChartViewer.vue
nicolaskempf57: I'll try if I can add the aggregate to the series name, bu I don't think we should spend too much time of an edge case of handling multiple series, that is already out of scope of this version

nicolaskempf57: I changed the series name. When dealing with multiple series, we will have to update the legend to better format the serie name, but we don't want the legend for 1 serie, so we can do it later

components/Charts/ChartConfigurator.vue
nicolaskempf57: I updated the `removeFilter` to handle the Filter case, even if the front-end was always creating AndFilter. This way, all changes for the OrFilter are in this file

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
nicolaskempf57: You're right, I forgot to remove the previous status update in fetchSeriesProfile and fetchSeriesData too.

I find it easier to read to have 2 functions with their own purpose.

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
```diff
@@ -0,0 +1,225 @@
...
+      .map(r => r.value)
+    series.data = Object.fromEntries(results.map(result => [
+      result.id,
+      result.data.data,
+    ]))
+
+    for (const result of results) {
+      const serie = props.chart.series.find(s => s.resource_id === result.id)
+      if (serie) {
+        series.page[result.id] = 0
```
ThibaudDauce: After the initial `fetchSeriesData()` loads page 1, the code resets `series.page[result.id] = 0`. When `loadAllPages` is true, `loadMorePages()` then computes `nextPage = (series.page[resourceId] || 0) + 1 = 1`, refetches page 1, and concatenates it to the existing data — so the first 100 rows appear twice.

components/Charts/ChartConfigurator.vue
```diff
@@ -0,0 +1,623 @@
...
+    query: { q, size: 5 },
+  })
+}
+
+function removeFilter(index: number) {
+  if (!form.value.filter) return
+  if (!('filters' in form.value.filter)) {
+    return form.value.filter = null
+  }
+  form.value.filter.filters.splice(index, 1)
```
ThibaudDauce: `filterList` (line 658) hides nested `AndFilters` via `.filter((f): f is Filter => 'column' in f)`, so the UI index is the index in the *filtered* array. But `removeFilter` (line 589) calls `splice(index, 1)` on the *unfiltered* `form.value.filter.filters`. If a nested `AndFilters` precedes a `Filter` in the real array, the wrong element gets deleted (typically the nested `AndFilters`, silently). Reachable when loading a chart whose `series[0].filters` contains nested filters — `toChartForm` assigns them as-is without flattening.

ThibaudDauce (approved):

datagouv-components/src/components/Chart/ChartViewerWrapper.vue
nicolaskempf57: I will remove the loadMore for now because we explicitly rule it out of scope

components/Charts/ChartConfigurator.vue
nicolaskempf57: switched to template filtering ! This logic will change with the handling of "filter group" in cdata

nicolaskempf57:
I'm waiting for demo crawler to pass on the resource used for both local testing and CI, and I think we'll good to merge it

PR #1012, opened 2026-03-25
feat: new custom swagger

ThibaudDauce:
image

image


maudetes (approved):
Let's merge as is and iterate later to maybe support HTML/Markdown

PR #1022, opened 2026-04-01
feat(datagouv-components): add listbox components

nicolaskempf57:
Component extracted from the chart PR, because it's not really related.

This is a select-like component without the select issues related to the default styles and other layout things.


nicolaskempf57:
Done in https://github.com/datagouv/cdata/pull/963

PR #1023, opened 2026-04-03
feat: Nuxt 4.0 (minimal)

ThibaudDauce:
Try a minimal update to Nuxt 4.0 without touching TypeScript and file structure.


nicolaskempf57 (approved):
why is the ts config change necessary ?

ThibaudDauce:
> why is the ts config change necessary ?

Because Nuxt 4.0 enable this option by default

PR #1025, opened 2026-04-07
feat: add more information on preview missing

ThibaudDauce:


nicolaskempf57 (approved):
Nice ! Could we provide a link to the download tab? It seems like a nice UX improvement but maybe not easy to do

ThibaudDauce:
> Nice ! Could we provide a link to the download tab? It seems like a nice UX improvement but maybe not easy to do

We could improve it again yes!

agarrone:
I'm not sure about this since the download is going to work as a dropdown later.

PR #1026, opened 2026-04-07
fix: oembeds link

ThibaudDauce:
When we click on an oembed card, the link is loaded inside the iframe instead of propagating to the parent.


nicolaskempf57 (approved):
Thanks for this ! Why is it required ?

PR #1027, opened 2026-04-07
fix: tweaks post with blocs

ThibaudDauce:
- [x] Remove stripes between even/odd blocs
- [x] Fix edit button position and label (we had two "Modify" buttons one for the metadata one for the inline modification of blocs)
- [x] Container changes
    - [x] Change post container to match the one of the header / menu
    - [x] Left align posts
    - [x] Add back container only on texts (allow cards to overflow)

image


nicolaskempf57 (approved):
Working great after the conflict fixes 👍🏻 I don't know if we want the post content left aligned or centered ?

maudetes:
I've created [this issue](https://github.com/datagouv/data.gouv.fr/issues/2004) in order to keep track of the migration, cc @Pierlou

PR #1050, opened 2026-04-16
fix: cell overflow in moderation dashboard

ThibaudDauce:
Not the best solution but it's really weird and truncation in CSS is hard to understand…


maudetes (approved):
Indeed, truncate may be better but this works

PR #1052, opened 2026-04-17
feat(activity): show for org members

nicolaskempf57:


components/TabLinks.vue
```diff
@@ -30,10 +32,16 @@ const isCurrentUrl = useIsCurrentUrl()
...
-    return me.value?.roles?.includes(route.meta.requiredRole as string) ?? false
+  const requiredOrganizationPermission = route.meta.requiredOrganizationPermission as (keyof OrganizationReference['permissions'] | undefined) ?? null
+  if (requiredOrganizationPermission) {
+    const { currentOrganization } = useCurrentOwned()
+    if (currentOrganization.value) {
+      return currentOrganization.value.permissions[requiredOrganizationPermission] ?? false
+    }
+
+    return false
   }
```
ThibaudDauce: Same here… Visiting http://dev.local:3000/admin/datasets/66a7aad2d369533b90b4eb38 with SSR is working (I see the Activities tab) but without SSR while navigating it seems that `currentOrganization` is `null` so I don't have the tab… Not sure why…

ThibaudDauce (commented):
Doesn't work on my machine but maybe not linked to this PR. Seems to have weird side effects with `currentOrganization`

middleware/auth.ts
```diff
@@ -1,16 +1,25 @@
...
-    throw createError({ statusCode: 401, statusMessage: 'Unauthorized' })
+  if (requiredOrganizationPermission) {
+    const { currentOrganization } = useCurrentOwned()
+
+    if (currentOrganization.value) {
+      const permissionValue = currentOrganization.value.permissions[requiredOrganizationPermission]
+      if (!permissionValue) {
+        throw createError({ statusCode: 401, statusMessage: 'Unauthorized' })
+      }
+    }
```
ThibaudDauce: Not sure this middleware is doing something. I can visit http://dev.local:3000/admin/datasets/6727201449625899ea32c0d4/activities without triggering (with the production backend, without being super-admin) because I think `currentOrganization.value` is `null`

maudetes:
As discussed, activities should show for all members (not only admin)

pages/admin/topics/[id]/activities.vue
```diff
@@ -5,10 +5,6 @@
 <script setup lang="ts">
 import { ActivityList, type TopicV2 } from '@datagouv/components-next'
 
-definePageMeta({
-  requiredRole: 'admin',
-})
```
Comment (ThibaudDauce) [inline pages/admin/topics/[id]/activities.vue:10]: Did you add something to prevent clicking on topics activities?

ThibaudDauce (approved):
Why not do `show: () => dataset.permissions?.edit` instead of using meta of the page?

And why `show` is a function and not a `computed` boolean or directly a boolean?

pages/admin/topics/[id]/activities.vue
Comment (nicolaskempf57) [reply pages/admin/topics/[id]/activities.vue:10]: No, but topics aren't shown for an organization, it's only for power-users or admins so I think we want them to access the activity page

nicolaskempf57:
> Why not do `show: () => dataset.permissions?.edit` instead of using meta of the page?
> 
> And why `show` is a function and not a `computed` boolean or directly a boolean?

Your right, with the latest changes to use a function I don't think the meta is useful now and we can access the permissions

PR #1054, opened 2026-04-20
feat(search): add organizations facet support

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1020

Adds `organization_facet` as a filter option for `GlobalSearch`, allowing downstream consumers to filter by organizations present in the current result set (facet-based), as opposed to the existing `organization` filter which queries all organizations via the suggestion API. 

New `OrganizationFacetFilter` component renders `organization_id_with_name` facets (already returned by the API) as a radio group. 

`SearchFilterAliases` maps UI-only filter keys to their API param equivalents (`organization_facet` → `organization`).

Usage:

```typescript
getDefaultDatasetConfig({
  basicFilters: ['organization_facet'],
})
```

### DIFF
--- a/assets/css/overrides.css
+++ b/assets/css/overrides.css
@@ -30,10 +30,6 @@
   @apply !bg-none underline underline-offset-4 hover:decoration-2 text-new-primary;
 }
 
-.input {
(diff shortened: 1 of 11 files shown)


abulte:
Problem: only top X (50?) organisations are returned, this is probably blocking on ecologie.data.gouv.fr's side.

PR #1056, opened 2026-04-21
fix: try to fix new tabular /explore page

ThibaudDauce:
https://demo.data.gouv.fr/explore is crashing right now. I don't understand why it works in local but not in dev/demo.

I simplified the `useFetch()` and filter the result list to be more easily testable. Don't want to spend too much time on a test page…

### DIFF
--- a/pages/explore.vue
+++ b/pages/explore.vue
@@ -19,7 +19,7 @@
 
       <form
         class="fr-search-bar fr-search-bar--lg w-full"
-        @submit.prevent="search"
+        @submit.prevent="() => execute()"
       >
         <label
           for="search-input"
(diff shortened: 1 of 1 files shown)


pages/explore.vue
```diff
@@ -123,25 +114,36 @@
...
 useSeoMeta({ robots: 'noindex' })
 
 const query = ref('')
 const resourceId = ref('')
-const searchQuery = ref('')
 
-const { data: results, status } = await useAPI<PaginatedArray<Dataset>>(
-  computed(() => searchQuery.value ? `/api/1/datasets/?q=${encodeURIComponent(searchQuery.value)}&page_size=10` : ''),
-  { lazy: true, server: false },
+const searchParams = computed(() => ({ q: query.value, page_size: 10 }))
```
nicolaskempf57: Can we add `format: 'csv'` to reduce the chance of getting "Aucun résultat" ?

pages/explore.vue
nicolaskempf57, line 14: Should we add a query param to set the resourceId too ? Or will it be a separate page later ?

pages/explore.vue
ThibaudDauce: Done in [4dde2c1](https://github.com/datagouv/cdata/pull/1056/commits/4dde2c1c38ccecebe06c1a4e85c69a8c0e2a929c)

pages/explore.vue
ThibaudDauce: Yes this is only a test page, will do something way better when we integrate this into the ressource accordion

nicolaskempf57 (approved):

PR #1062, opened 2026-04-29
feat(search): export more types

### LINKED ISSUE #1083: datagouv/components: export more types
Two missing exports from @datagouv/components-next

  src/main.ts exports getDefaultDatasetConfig / getDefaultDataserviceConfig /
  getDefaultTopicConfig and GlobalSearchConfig, but not the underlying config or filter types
   those functions operate on. Downstream consumers that build GlobalSearchConfig
  programmatically need to type their override objects, which currently forces workarounds.

  1. Per-class config and filter types

  Not exported: DatasetSearchConfig, DataserviceSearchConfig, TopicSearchConfig,
  HiddenFilter<T>, DatasetSearchFilters, DataserviceSearchFilters, TopicSearchFilters.

  Without these, the only way to type an overrides object is:

  // current workaround — fragile if function signature changes
  baseArgs as Parameters<typeof getDefaultDatasetConfig>[0]

  With them, callers could write:

  import type { DatasetSearchConfig, HiddenFilter, DatasetSearchFilters } from
  '@datagouv/components-next'

  const overrides: Omit<DatasetSearchConfig, 'class'> = {
    hiddenFilters: [{ key: 'organization', value: '...' }] satisfies
  HiddenFilter<DatasetSearchFilters>[],
    basicFilters: ['tag'],
  }

  2. Native filter key union

  Not exported: a type or enum for the set of filter keys valid in
  basicFilters/advancedFilters. Downstream code that reads filter config from YAML needs to
  distinguish native filter types (go into basicFilters) from custom ones (rendered in
  #custom-filters slot). Currently we maintain a local enum mirroring the upstream set:

  // local workaround — silently stale if upstream adds a filter type
  enum GlobalSearchNativeFilterType {
    Tag = 'tag', Organization = 'organization', Badge = 'badge', /* … */
  }

  Exporting a union like type NativeFilterKey = keyof DatasetSearchFilters | keyof
  DataserviceSearchFilters | keyof TopicSearchFilters (or a Set/enum equivalent) would remove
   the need for this local copy.

  Ask: add the types above to the src/main.ts re-export list.

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1083

### DIFF
--- a/datagouv-components/src/types/search.ts
+++ b/datagouv-components/src/types/search.ts
@@ -351,6 +351,8 @@ export type TopicSearchConfig = {
 
 export type SearchTypeConfig = DatasetSearchConfig | DataserviceSearchConfig | ReuseSearchConfig | OrganizationSearchConfig | TopicSearchConfig
 
+export type BuiltInFilterKey = keyof DatasetSearchFilters | keyof DataserviceSearchFilters | keyof TopicSearchFilters
+
 export type SearchType = SearchTypeConfig['class']
 
 export type GlobalSearchConfig = SearchTypeConfig[]
--- a/datagouv-components/src/types/search.ts
+++ b/datagouv-components/src/types/search.ts
@@ -351,7 +351,7 @@ export type TopicSearchConfig = {
 
 export type SearchTypeConfig = DatasetSearchConfig | DataserviceSearchConfig | ReuseSearchConfig | OrganizationSearchConfig | TopicSearchConfig
 
-export type BuiltInFilterKey = keyof DatasetSearchFilters | keyof DataserviceSearchFilters | keyof TopicSearchFilters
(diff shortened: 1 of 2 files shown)


datagouv-components/src/types/search.ts
```diff
@@ -351,6 +351,8 @@ export type TopicSearchConfig = {
 
 export type SearchTypeConfig = DatasetSearchConfig | DataserviceSearchConfig | ReuseSearchConfig | OrganizationSearchConfig | TopicSearchConfig
 
+export type BuiltInFilterKey = keyof DatasetSearchFilters | keyof DataserviceSearchFilters | keyof TopicSearchFilters
```
ThibaudDauce: Why no `ReuseSearchFilters` and `OrganizationSearchFilters`?

abulte: Because I don't use them 😅 But yeah I'll add them 👍

ThibaudDauce (approved):

PR #1067, opened 2026-05-01
feat(search): type aware custom filters

### LINKED ISSUE #1089: datagouv/components: Type-scoped custom filters in GlobalSearch
Brief: Type-scoped custom filters in GlobalSearch

  Problem

  customFilterRegistry is a flat Map<string, CustomFilterEntry> shared across all type
  searches. useStableQueryParams calls forEachActiveCustomFilter with no type guard, so every
   registered custom filter is injected into every type's API params. This causes background
  searches for non-active types to be polluted by filters that only apply to the current
  type.

  Clearing on type switch works today only because SearchSelectFilter unmounts (new route,
  new customFilters meta), triggering onScopeDispose → unregister → registry empties.
  GlobalSearch has no actual type awareness.

  ---
  Proposed changes

  1. CustomFilterEntry — add optional typeKeys

  export interface CustomFilterEntry {
    apiParam: string
    ref: Ref<string | undefined>
    defaultValue: string | undefined
    typeKeys?: string[]   // undefined = applies to all types (current behavior)
  }

  2. UseSearchFilterOptions — add optional typeKeys

  export interface UseSearchFilterOptions {
    apiParam?: string
    defaultValue?: string
    typeKeys?: string | string[]  // one or more type config keys this filter applies to
  }

  3. useSearchFilter — forward typeKeys on register

  const { apiParam = urlParam, defaultValue = undefined, typeKeys } = options
  const normalizedTypeKeys = typeKeys
    ? (Array.isArray(typeKeys) ? typeKeys : [typeKeys])
    : undefined

  onMounted(() => {
    context.register(urlParam, { apiParam, ref: value, defaultValue, typeKeys:
  normalizedTypeKeys })
  })

  4. forEachActiveCustomFilter — add optional typeKey filter arg

  export function forEachActiveCustomFilter(
    registry: Map<string, CustomFilterEntry>,
    apply: (apiParam: string, value: string) => void,
    typeKey?: string,           // if provided, skip entries that don't include it
  ): void {
    for (const entry of registry.values()) {
      if (!isCustomFilterActive(entry)) continue
      if (typeKey && entry.typeKeys && !entry.typeKeys.includes(typeKey)) continue
      apply(entry.apiParam, String(entry.ref.value))
    }
  }

  The double condition (typeKey && entry.typeKeys) means:
  - no typeKey arg → all entries apply (call-site backward compat)
  - entry has no typeKeys → applies to all types (registration backward compat)

  5. useStableQueryParams — pass typeConfig.key when iterating custom filters

  forEachActiveCustomFilter(customFilterRegistry, (apiParam, value) => {
    const existing = params[apiParam]
    params[apiParam] = existing === undefined
      ? value
      : Array.isArray(existing) ? [...existing, value] : [existing, value]
  }, typeConfig?.key)   // ← new arg

  Same change needed in the RSS URL builder in GlobalSearch.vue (the other
  forEachActiveCustomFilter call) — pass currentTypeConfig.value?.key.

  ---
  Backward compatibility

  No breaking changes. All three call paths degrade gracefully when the new fields are
  absent:
  - register without typeKeys → entry has typeKeys: undefined → applies to all types
  - forEachActiveCustomFilter without third arg → typeKey is undefined → no filtering

  ---
  Downstream usage (ecospheres)

  SearchSelectFilter receives the current pageKey and passes it as typeKeys:

  const urlValue = useSearchFilter(props.config.urlParam, {
    apiParam: props.config.apiParam,
    typeKeys: props.config.typeKeys,   // string[] from SelectFilterConfig
  })

  SelectFilterConfig grows a typeKeys: string[] field, populated in buildGlobalSearchConfig
  with [pageKey]. A filter shared across two page types would list both keys.

abulte:
Fix https://github.com/ecolabdata/ecospheres/issues/1089

This introduces a new config parameter `typeKeys` that lets downstream say if a custom filter is shared across types or not. If it's not shared, the background search for other types is skipped. This behaves similarly to native filters.

### DIFF
--- a/datagouv-components/src/components/Search/GlobalSearch.vue
+++ b/datagouv-components/src/components/Search/GlobalSearch.vue
@@ -356,7 +356,7 @@ import { RiBookShelfLine, RiBuilding2Line, RiCloseCircleLine, RiDatabase2Line, R
 import magnifyingGlassSrc from '../../../assets/illustrations/magnifying_glass.svg?url'
 import { useTranslation } from '../../composables/useTranslation'
 import { useDebouncedRef } from '../../composables/useDebouncedRef'
-import { forEachActiveCustomFilter, isCustomFilterActive, searchFilterContextKey, type CustomFilterEntry } from '../../composables/useSearchFilter'
+import { configKey, forEachActiveCustomFilter, isCustomFilterActive, searchFilterContextKey, type CustomFilterEntry } from '../../composables/useSearchFilter'
 import { useStableQueryParams } from '../../composables/useStableQueryParams'
 import { useComponentsConfig } from '../../config'
 import { useFetch } from '../../functions/api'
@@ -366,7 +366,7 @@ import type { Dataservice } from '../../types/dataservices'
 import type { Organization } from '../../types/organizations'
 import type { Reuse } from '../../types/reuses'
 import type { TopicV2 } from '../../types/topics'
-import type { GlobalSearchConfig, SearchResponseByClass, SearchType, SearchTypeConfig, SortOption, FacetItem } from '../../types/search'
+import type { GlobalSearchConfig, SearchResponseByClass, SearchType, SortOption, FacetItem } from '../../types/search'
 import { getDefaultGlobalSearchConfig } from '../../types/search'
 import BrandedButton from '../BrandedButton.vue'
 import LoadingBlock from '../LoadingBlock.vue'
@@ -405,8 +405,6 @@ const props = withDefaults(defineProps<{
   config: getDefaultGlobalSearchConfig,
   hideSearchInput: false,
 })
-
-const configKey = (c: SearchTypeConfig) => c.key ?? c.class
 
 // defineModel's default is static and can't depend on props, so we cast and initialize manually
 const currentType = defineModel<string>('type') as Ref<string>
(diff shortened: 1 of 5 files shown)


datagouv-components/src/components/Search/GlobalSearch.vue
```diff
@@ -742,7 +742,7 @@ const rssUrl = computed(() => {
 
   forEachActiveCustomFilter(customFilterRegistry, (apiParam, value) => {
     params.set(apiParam, value)
-  })
+  }, currentTypeConfig.value?.key ?? currentTypeConfig.value?.class)
```
ThibaudDauce: Can you use the `configKey()` helper?

datagouv-components/src/composables/useStableQueryParams.ts
```diff
@@ -55,6 +55,9 @@ export function useStableQueryParams(options: StableQueryParamsOptions) {
     // 3.5. Apply custom filter values. Concatenate into an array on collision
     // so a custom filter mapped onto a built-in apiParam (e.g. theme → tag)
     // combines with an existing built-in value instead of overwriting it.
+    // Pass the current type key so filters scoped to specific types are excluded
+    // from background fetches for other types.
+    const currentTypeKey = typeConfig?.key ?? typeConfig?.class
```
ThibaudDauce: Maybe refactor `configKey` outside the component to use it here too?

ThibaudDauce (approved):
Not a fan of the duplication between the v-if and the `useSearchFilter` but I don't have a better solution right now :-)

PR #1074, opened 2026-05-07
feat: use new harvest API

ThibaudDauce:
- Require https://github.com/opendatateam/udata/pull/3759


nicolaskempf57 (approved):
Seems to be working great !

PR #1075, opened 2026-05-07
feat: enhance contact display

### LINKED ISSUE #1907: Mise en avant du contact et de la demande d'accès dans une fiche API
On a pas mal de demandes de support cherchant à contacter les producteurs d'une API spécifique.

## Le contact aujourd'hui

Image

## Ce qu'on aimerait

Avoir un bouton, ou un lien, plus visible, mieux mis en avant.

Samuelfaure:
fix https://github.com/datagouv/data.gouv.fr/issues/1907

Screens:

screen1
screen2

### DIFF
--- a/components/AccessTypes/AccessTypePanel.vue
+++ b/components/AccessTypes/AccessTypePanel.vue
@@ -1,6 +1,9 @@
 <template>
   <div class="space-y-1">
-    <dt class="text-gray-plain font-bold">
+    <dt
+      class="text-gray-plain font-bold"
+      :class="{ 'sr-only': hideAccessLabel }"
+    >
       {{ $t('Accès') }}
     </dt>
     <dd class="p-0">
@@ -11,19 +14,6 @@
       <template v-else>
         {{ $t('Non spécifié') }}
       </template>
-      <div
-        v-if="object.authorization_request_url"
-        class="mt-2.5"
-      >
-        <BrandedButton
-          color="primary"
-          size="sm"
-          :href="object.authorization_request_url"
-          new-tab
-        >
-          {{ $t("Demander un accès") }}
-        </BrandedButton>
-      </div>
     </dd>
     <template v-if="object.access_type === 'restricted'">
       <dt class="text-gray-plain font-bold mt-2.5">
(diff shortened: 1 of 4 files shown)


pages/dataservices/[did].vue
```diff
@@ -84,132 +84,147 @@
                 />
               </ReadMore>
             </div>
-            <dl
+            <aside
```
Comment (ThibaudDauce) [inline pages/dataservices/[did].vue:87]: This broke the `dt` and `dd` below… `dl` should stay?

components/AccessTypes/AccessTypePanel.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="space-y-1">
-    <dt class="text-gray-plain font-bold">
+    <dt
+      v-if="!hideAccessLabel"
```
ThibaudDauce: A `dd` without a `dt` is invalid. Maybe use a `sr-only` class instead?

tests/dataservices/[did].spec.ts
```diff
@@ -13,18 +13,34 @@ test('page loads with correct title', async ({ page }) => {
...
+  await expect(page.getByText('Date de création')).toBeVisible()
   await expect(page.getByText('Taux de disponibilité')).toBeVisible()
   await expect(page.getByText('Non communiqué')).toBeVisible()
 })
 
+test('sidebar Swagger button reveals the viewer', async ({ page }) => {
+  await page.goto(`/dataservices/${DATASERVICE_SLUG}`)
+  await page.waitForLoadState('networkidle')
+
+  const sidebarSwagger = page.getByRole('button', { name: 'Swagger' }).first()
```
Comment (ThibaudDauce) [inline tests/dataservices/[did].spec.ts]: Maybe use something like `page.locator('aside').getByRole('button', { name: 'Swagger' })` otherwise a change in the HTML order could lead to testing the "real" Swagger button and not the button in the sidebar.

pages/dataservices/[did].vue
Comment (Samuelfaure) [reply pages/dataservices/[did].vue:87]: I kept `<aside>` as the outer (h3/section/button-group can't live inside a `<dl>`) and instead wrapped each dt/dd group in its own `<dl>`

ThibaudDauce (approved):

PR #1076, opened 2026-05-11
feat(notifications): add new notifications

nicolaskempf57:


components/Notifications/NewDataservice.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+  <div class="p-3 flex gap-3 relative hover:bg-gray-some">
+    <div class="flex-none">
+      <RiTerminalLine class="size-4" />
+    </div>
+    <div class="flex-1 truncate">
+      <p class="m-0 text-xs font-bold">
+        <CdataLink
+          class="after:absolute after:inset-0 bg-none"
```
ThibaudDauce: There is two `after:absolute after:inset-0` in this component, is there a conflict between them?

ThibaudDauce (approved):
Still not a big fan of all this duplication between notifications. Maybe we should create a NotificationLayout component and expose `title`, `icon`, etc. to be more lean.

components/Notifications/NewDataservice.vue
nicolaskempf57: The button is unecessary

nicolaskempf57:
> Still not a big fan of all this duplication between notifications. Maybe we should create a NotificationLayout component and expose `title`, `icon`, etc. to be more lean.

Done

components/Notifications/NotificationLayout.vue
```diff
@@ -0,0 +1,66 @@
+<template>
+  <div class="p-3 flex gap-3 relative hover:bg-gray-some">
+    <div class="flex-none">
+      <component
+        :is="icon"
+        class="size-4"
+      />
+    </div>
+    <div class="flex-1 truncate">
+      <p class="m-0 text-xs font-bold relative">
```
ThibaudDauce: The `relative` class on this `<p>` scopes the `CdataLink`'s `after:absolute after:inset-0` overlay to the title paragraph instead of the outer notification card. Previously every notification was fully clickable. With this change only the title text triggers navigation — clicks on the date, slot content (message preview, `OrganizationOwner`, etc.) no longer navigate. The pre-refactor components (`Discussion.vue`, `ValidateHarvester.vue`, `MembershipRequest.vue`, `TransferRequest.vue`) all had a non-positioned `<p>` precisely so the overlay extended to the outer `relative` `<div>`.

```suggestion
      <p class="m-0 text-xs font-bold">
```

components/Notifications/NotificationLayout.vue
```diff
@@ -0,0 +1,66 @@
...
+        <CdataLink
+          v-if="titleLink"
+          class="after:absolute after:inset-0 bg-none truncate"
+          :to="titleLink"
+          :title="titleLinkTitle"
+          @click="markAsRead(notification)"
```
ThibaudDauce: This unconditionally marks the notification as read on title-link click. Three of the migrated components intentionally avoided this for action-required notifications:

- `ValidateHarvester` had a `handleLinkClick` that skipped `markAsRead` when `status === 'pending'`.
- `MembershipRequest` used `<NuxtLink>` without `@click`, so clicking the link did not mark it read.
- `TransferRequest` used `<NuxtLink>` without `@click`, same behavior.

These notifications had a red `bg-danger` bullet and were only cleared once the user accepted or refused. After the refactor a simple click marks them read even without taking action. If this is intentional please confirm, otherwise the `markAsRead` call should be gated on `!requiresAction` (or moved back to each call-site).

components/Notifications/NotificationLayout.vue
```diff
@@ -0,0 +1,66 @@
...
+  </div>
+</template>
+
+<script setup lang="ts">
+import { AnimatedLoader, useFormatDate } from '@datagouv/components-next'
+import CdataLink from '../CdataLink.vue'
+import type { UserNotification } from '~/types/notifications'
+
+defineProps<{
+  icon: object
```
ThibaudDauce: The icon prop type should be `Component` from `vue` to match the codebase convention (`components/BannerNotif.vue`, `datagouv-components/src/components/Tag.vue`, `BrandedButton.vue`, `ObjectCardBadge.vue`, `ObjectCardHeader.vue`, `RadioInput.vue`).

```suggestion
  icon: Component
```

And add at the top of the script block: `import type { Component } from 'vue'`.

components/Notifications/Discussion.vue
```diff
@@ -1,88 +1,61 @@
...
+        :with-link="false"
       />
-      <div
-        v-else-if="!notification.handled_at"
-        class="size-2 rounded-full bg-new-primary mt-0.5"
-      />
-    </div>
-  </div>
+    </p>
+    <p class="m-0 text-xs truncate">
```
ThibaudDauce: Before the refactor the message content (`message.content`) was inside a `<CdataLink>` with `:class="{ 'text-gray-medium': notification.details.status === 'new_comment' }"`. The conditional gray styling for new-comment messages is lost in the slot version — the `<p>` here has no equivalent class, so new-comment notifications look identical to new-discussion ones.

```suggestion
    <p
      class="m-0 text-xs truncate"
      :class="{ 'text-gray-medium': notification.details.status === 'new_comment' }"
    >
```

ThibaudDauce (approved):

PR #1081, opened 2026-05-18
fix: hide rotate button when already asked

ThibaudDauce:


components/User/AdminUserProfilePage.vue
```diff
@@ -179,7 +179,10 @@
           {{ $t("L'utilisateur sera déconnecté et devra définir un nouveau mot de passe à sa prochaine connexion.") }}
         </template>
 
-        <template #button>
+        <template
+          v-if="!rotationPending"
```
maudetes, line 183: Don't we prefer a `disabled` behavior?

maudetes (approved):

PR #1084, opened 2026-05-19
chore: add timestamp in logging

maudetes:
Part of https://github.com/datagouv/data.gouv.fr/issues/2015

We don't currently have any timestamp in logging, this PR prefix logs with a timestamp.
To raise an error, go to: http://dev.local:3000/users/does-not-exist

I think this solution is ugly, but at the moment I did not find a good approach that don't require adding an extra dependency.
Any other better idea?


plugins/logger.ts
```diff
@@ -0,0 +1,25 @@
+import { defineNuxtPlugin } from '#imports'
+
+// Module-level variable to ensure one-time initialization
+let __logger_plugin_initialized = false
+
+export default defineNuxtPlugin(() => {
+  if (process.env.NODE_ENV === 'production' && !__logger_plugin_initialized) {
```
ThibaudDauce: I prefer early returns for this kind of stuff but not mandatory:



```suggestion
  if (__logger_plugin_initialized) return
  if (process.env.NODE_ENV !== 'production') return
```

ThibaudDauce (approved):

PR #1088, opened 2026-05-20
Feat/charts improvements

### LINKED ISSUE #2014: Datastudio : premiers quickfixes
Fait suite à https://github.com/datagouv/data.gouv.fr/issues/1972.

Premiers retours sur la version en cours de dev :

- [x] Renommer Série en 'Axe Y'
- [x] Déplacer les styles de l'axe dans un accordéon
- [x] Masquer la légende
- [x] Changer la font utilisée par echart pour la Marianne et le bleu `#000091`
- [x] Retirer le titre d'echart
- [x] Afficher le titre et la description en texte non éditable au dessus du chart
- [x] Ajouter un séparateur horizontal entre titre + desc et la previz en dessous
- [x] Générer la source avec : nom du jdd, nom de l'orga, datagouv
- [x] Afficher un blank state
- [x] Masquer les inputs à droite tant qu'on n'a pas un jeu de données et une ressource sélectionnée
- [x] Avec qui souhaitez publier avant la partie donnée ? ou tout à la fin ?
- [x] Afficher la barre de droite de la meme taille que l'aide
- [x] supprimer "choix de la ressource" si aucun jeu de données est sélectionné
- [x] vérifier si "jeu de données" est scopé au niveau de l'orga
- [x] mettre "choisir quoi afficher" à la place de column dans axe X
- [x] Column : Ajouter une recherche et afficher un icône correspondant aux types de colonnes
- [x] colonne : classer les colonnes par type et ajouter un séparateur entre chaque
- [x] trier : ajouter icone pour indiquer le sens (juste une flèche)
- [x] Retirer les dates de l'axe Y
- [x] Ajouter un icone dans le sélecteur pour Type de graphique
- [x] Retirer les ressources non apifiées sur sélecteur de ressource

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/2014


components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +742,109 @@ async function saveChart() {
...
   }
 }
+
+watch(
+  () => ({
+    series: form.value.series,
+    x_axis: form.value.x_axis,
+    y_axis: form.value.y_axis,
+    filter: form.value.filter,
+    extras: form.value.extras,
```
ThibaudDauce, line 862: Do we need to watch for `chart_type`?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
+const chartTypeProxy = computed<{ value: string, label: string, icon: Component } | null>({
+  get: () => chartTypeOptions.find(o => o.value === form.value.chart_type) ?? null,
+  set: (val: { value: string, label: string, icon: Component } | null) => {
+    if (val) {
+      form.value.chart_type = val.value as DataSeriesType
+    }
+  },
+})
+
+const xAxisColumnProxy = computed<{ name: string, colType: ColumnType, resourceId: string } | null>({
```
ThibaudDauce: `{ name: string, colType: ColumnType, resourceId: string }` is repeated a few times, maybe named it?

components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +742,109 @@ async function saveChart() {
...
+  if (r) {
+    await loadMissingResourcesForChart(new Set([r]))
+    if (columns.value[r]?.length > 0) {
+      return
+    }
+    const profile = await getProfile(r)
+    resourceProfiles.value[r] = {
+      profile: { categorical: profile.profile.categorical },
+      columns: profile.profile.columns,
+    }
```
ThibaudDauce: Never used?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
+              <label
+                for="x-axis-sort-combined"
+                class="fr-label"
+              >{{ $t('Trier par') }}</label>
+              <Listbox
+                id="x-axis-sort-combined"
+                v-model="sortProxy"
+                class="w-full"
+                :options="sortOptions"
+                :display-value="(opt) => opt ? `${opt.column}${opt.column ? ' - ' : ''}${opt.label}` : $t('Sélectionnez une option')"
```
ThibaudDauce: `${opt.column}${opt.column ? ' - ' : ''}${opt.label}` is repeated a few times? With `sortProxy` or `option` below. Maybe extract an helper?

components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +742,109 @@ async function saveChart() {
...
+    resourceProfiles.value[r] = {
+      profile: { categorical: profile.profile.categorical },
+      columns: profile.profile.columns,
+    }
+    columns.value[r] = profile.profile.header.map((name) => {
+      const colInfo = profile.profile.columns[name]
+      const isCategorical = profile.profile.categorical.includes(name)
+      const colType = resolveColumnType(colInfo ?? { python_type: 'unknown', format: undefined }, isCategorical)
+      return { name, type: colType }
+    })
```
ThibaudDauce: This map is repeated below with `result.profile.profile.header.map`

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
 import { useAPI } from '~/utils/api'
-
+import { isMeAdmin } from '~/utils/auth'
 import ChartFilterRow from './ChartFilterRow.vue'
+import ProducerSelect from '../ProducerSelect.vue'
+import Accordion from '~/components/Accordion/Accordion.global.vue'
+import AccordionGroup from '~/components/Accordion/AccordionGroup.global.vue'
 import type { DatasetSuggest } from '~/types/types'
 
+type CombinedSort = '' | 'axis_x-asc' | 'axis_x-desc' | 'axis_y-asc' | 'axis_y-desc'
```
ThibaudDauce: Already exists in `visualizations.ts`

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
-  if (!newDataset) return
-  try {
-    const fetchedResources = await $chartsApi<PaginatedArray<Resource>>(`/api/2/datasets/${newDataset.id}/resources/`)
-    for (const r of fetchedResources.data.filter(resource => hasTabularData(resource))) {
-      savedResources[r.id] = r
-    }
+  const filter = form.value.filter
+  if (isFilter(filter)) {
+    return [filter]
   }
```
ThibaudDauce, line 636: Could be a `return filter.filters.filter(isFilter)`?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
+})
+
+const xAxisColumnOptions = computed<Array<{ name: string, colType: ColumnType, resourceId: string }>>(() => {
+  const result: Array<{ name: string, colType: ColumnType, resourceId: string }> = []
+  for (const [resourceId, cols] of Object.entries(columns.value)) {
+    for (const col of cols) {
+      result.push({ name: col.name, colType: col.type, resourceId })
+    }
+  }
+  return result
```
ThibaudDauce: Could be a `Object.entries(...).flatMap(...)`?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,518 @@
...
+})
+
+const yAxisColumnOptions = computed<Array<{ name: string, colType: ColumnType }>>(() => {
+  const result: Array<{ name: string, colType: ColumnType }> = []
+  if (selectedResource.value && columns.value[selectedResource.value]) {
+    for (const col of columns.value[selectedResource.value]) {
+      if (col.type !== 'date') {
+        result.push({ name: col.name, colType: col.type })
+      }
+    }
```
ThibaudDauce: Could be `.filter().map()` ?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,525 @@
...
+                    {{ $t('Maximum') }}
+                  </option>
+                  <option value="median">
+                    {{ $t('Médiane') }}
+                  </option>
+                </select>
+              </div>
+            </div>
+            <AccordionGroup>
+              <Accordion title="Styles">
```
ThibaudDauce:
```suggestion
              <Accordion title="$t('Styles')">
```

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,525 @@
...
   required: true,
 })
 
-const { t } = useTranslation()
-
-const columns = ref<Record<string, Array<string>>>({})
-const columnsArray = computed(() => Object.entries(columns.value).map(([key, value]) => ({ key, value })))
-
-const dataset = ref<DatasetSuggest>({ id: '6170ae10981edd7b132f28a0', title: 'Logements et logements sociaux dans les EPCI' } as DatasetSuggest)
+const columns = ref<Record<string, Array<{ name: string, type: ColumnType }>>>({})
```
ThibaudDauce: This type is repeated 4 times?

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,525 @@
...
+  return yAxisColumnOptions.value.find(opt => opt.name === colName) ?? null
+}
+
+function buildColumnsFromProfile(profile: { profile: { header: string[], columns: Record<string, { python_type: string, format?: string }>, categorical: string[] } }): Array<{ name: string, type: ColumnType }> {
+  return profile.profile.header.map((name) => {
+    const colInfo = profile.profile.columns[name]
+    const isCategorical = profile.profile.categorical.includes(name)
+    const colType = resolveColumnType(colInfo ?? { python_type: 'unknown', format: undefined }, isCategorical)
+    return { name, type: colType }
+  })
```
ThibaudDauce: This header.map() is repeated twice?

components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +735,101 @@ async function saveChart() {
...
+    console.error('Failed to fetch resources:', error)
+  }
+}, { immediate: true })
+
+watch(producer, () => {
+  dataset.value = null
+  selectedResource.value = null
+})
+
+watch(selectedResource, async (r) => {
```
ThibaudDauce, line 887: The selected resource is never written into `form.series[].resource_id`: this watch only loads the profile and columns, and no other code writes `resource_id`. Since the page initializes `series[0].resource_id` with the hardcoded `14dba482-…`, both the preview (`ChartViewerWrapper` fetches profile and data from `chart.series[].resource_id`) and the saved payload (`toChartApi` spreads the series as-is) always point to the hardcoded resource, whatever dataset/resource the user picks. The e2e tests don't catch this because `setupChart` selects exactly that dataset.

Timing-dependent side effect: the wrapper's `@columns` emit rebuilds the whole `columns` object from `chart.series` resource ids only, so it can wipe the entry loaded here for the actually-selected resource and leave the "Colonne Y" options empty.

components/Charts/ChartConfigurator.vue
```diff
@@ -82,7 +120,18 @@
             </button>
           </div>
         </fieldset>
-        <fieldset class="px-6">
+        <fieldset class="min-w-0">
+          <ProducerSelect
+            v-model="producer"
```
ThibaudDauce, line 124: `producer` is never synced into `form.owned`: it's only used for `sourceText`, the `suggestDataset` query and the reset watcher. `saveChart()` sends `toChartApi(form)` which spreads `chartForm.owned`, still holding the page's default (`{ organization: null, owner: me.id }`), so the chart is always published as the current user regardless of the answer to "Avec qui souhaitez-vous publier ?". The other publication forms bind `ProducerSelect` to `form.owned` (e.g. `DescribeReuse.vue`, `DescribeDataset.vue`) — the same propagation is needed here (mapping the selected `Owned` to ids).

components/Charts/ChartConfigurator.vue
```diff
@@ -539,8 +691,6 @@ async function loadChart(id: string) {
       }
 
       form.value = toChartForm(data)
```
ThibaudDauce, line 742: Loading an existing chart now leaves the screen on the blank state: the left panel and all the config fieldsets (including the save button) are under `v-if="selectedResource"` / `v-if="producer"`, but `loadChart` never sets `producer`, `dataset` or `selectedResource`. After clicking "Charger", the user only gets the success toast — the chart is not displayed and cannot be edited or re-saved (regression: on main the panels were unconditional). `loadChart` should also restore `producer`, `dataset` and `selectedResource` from the loaded chart.

components/Charts/ChartConfigurator.vue
```diff
@@ -96,414 +145,511 @@
...
-const columns = ref<Record<string, Array<string>>>({})
-const columnsArray = computed(() => Object.entries(columns.value).map(([key, value]) => ({ key, value })))
-
-const dataset = ref<DatasetSuggest>({ id: '6170ae10981edd7b132f28a0', title: 'Logements et logements sociaux dans les EPCI' } as DatasetSuggest)
+const columns = ref<ColumnsDefinition>({})
+const producer = ref<Owned | null>(null)
+const dataset = ref<DatasetSuggest | null>(null)
 const savedResources = reactive<Record<string, Resource>>({})
-const selectedResource = ref('14dba482-41e3-4c54-b82a-d8c11d1d80eb')
+const selectedResource = ref<string | null>(null)
```
ThibaudDauce: With `null` here (and in the resets in `watch(producer)` / `watch(dataset)`), the placeholder of `#resource-select` never shows: the placeholder option has `value=""`, and Vue's `vModelSelect` finds no option loosely equal to `null` so it sets `selectedIndex = -1` — the select renders empty instead of showing "Sélectionnez une ressource". Initialize and reset with `''` instead of `null` (or use `:value="null"` on the placeholder option).

ThibaudDauce (changes requested):
A few bugs in the new publication flow — details inline.

components/Charts/ChartConfigurator.vue
nicolaskempf57: Nice catch ! With the issues with demo tabular, I didn't try other resource and I somehow broke the assignment

ThibaudDauce: Thanks Claude :-)

components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +811,102 @@ async function saveChart() {
...
+    for (const r of fetchedResources.data.filter(resource => hasTabularData(resource))) {
+      savedResources[r.id] = r
+    }
+  }
+  catch (error) {
+    console.error('Failed to fetch resources:', error)
+  }
+}, { immediate: true })
+
+watch(selectedResource, async (r) => {
```
ThibaudDauce, line 887: When the user switches resource (or switches dataset then resource), this watch updates `series[0].resource_id` and loads the new resource's columns, then calls `ensureSeriesHasColumnX/Y`. But those guards only fill the value when it's empty (`if (!column_x)`, `if (column_y === '')`). After the first auto-fill (or after a `loadChart`), `column_x`/`column_y` still hold column names from the *previous* resource and are never reset.

Result: `resource_id` now points to resource B while `column_x`/`column_y` still name columns of resource A. The "Colonne Y" `SearchableSelect` renders empty (`getSerieColumnYValue` finds nothing in `yAxisColumnOptions`, which is scoped to the selected resource) while `series[0].column_y` silently keeps the stale value, and `fetchSeriesData` then queries resource B with a column it doesn't have → broken preview/save.

The "y-axis column options update when selecting different resource" test only asserts that the option list changes, not that `column_y` is re-validated, so it doesn't catch this. The fix should reset (or re-validate against the new resource's columns) `column_x` and `series[0].column_y` whenever the resource changes, instead of only filling them when empty.

components/Charts/ChartConfigurator.vue
```diff
@@ -585,39 +811,119 @@ async function saveChart() {
...
+
+watch(dataset, async (newDataset) => {
+  selectedResource.value = ''
+  if (!newDataset) {
+    return
+  }
+  try {
+    const fetchedResources = await $chartsApi<PaginatedArray<Resource>>(`/api/2/datasets/${newDataset.id}/resources/`)
+    for (const r of fetchedResources.data.filter(resource => hasTabularData(resource))) {
+      savedResources[r.id] = r
```
ThibaudDauce, line 879: `savedResources` is never cleaned when changing dataset? We accumulate the old ressources with the new ones? I think `columns` has the same problem. `loadChart` accumulate too inside `savedResources` without cleaning?

ThibaudDauce (changes requested):

components/Charts/ChartConfigurator.vue
nicolaskempf57: Yes it was done this way at first to handle multiple series from different datasets / resources and later to avoid duplicated requests if the resource is already loaded.
We might want to simply this and not save previous resources / columns.

nicolaskempf57: Can we handle it in a subsequent PR ? We would have more time to test a 2-series chart made from the API for example

ThibaudDauce: Yes let's merge this :-) It's too big of a change and it has wait long enough… We can fix bugs later :-)

ThibaudDauce (approved):

PR #1095, opened 2026-05-25
chore(charts): make echarts optional 

abulte:
Make `echarts` optional by declaring it as a peer dependency and using a dedicated entry point for its usage. Two benefits:

1. Downstream that doesn't need the echarts components does not install the huge library
2. Avoid downstream issues for local install of components. The chart components imports in `main.ts` will trigger vue deps analysis until `echarts` is reached, and it might not be installed (it's not in front-kit since we don't need it yet).

```
[plugin:vite:import-analysis] Failed to resolve import "echarts/core" from "node_modules/@datagouv/components-next/src/components/Chart/ChartViewer.vue". Does the file exist?
/Users/alexandre/Developer/Ecolab/Ecopshères/ecospheres-front/node_modules/@datagouv/components-next/src/components/Chart/ChartViewer.vue:9:48
12 |  /* Injection by vite-plugin-vue-inspector End */
13 |  import { defineComponent as _defineComponent } from "vue";
14 |  import { format, use } from "echarts/core";
   |                               ^
15 |  import { CanvasRenderer } from "echarts/renderers";
16 |  import { LineChart, BarChart } from "echarts/charts";
```

I went with a non-optional peer dep so that downstream gets a non-blocking warning inviting to install echarts if needed.

### DIFF
--- a/components/Charts/ChartConfigurator.vue
+++ b/components/Charts/ChartConfigurator.vue
@@ -399,7 +399,7 @@ import { useAPI } from '~/utils/api'
 import ChartFilterRow from './ChartFilterRow.vue'
 import type { DatasetSuggest } from '~/types/types'
 
-const ChartViewerWrapper = defineAsyncComponent(() => import('@datagouv/components-next').then(m => m.ChartViewerWrapper))
+const ChartViewerWrapper = defineAsyncComponent(() => import('@datagouv/components-next/chart').then(m => m.ChartViewerWrapper))
 
 const form = defineModel<ChartForm>({
   required: true,
--- a/datagouv-components/package.json
+++ b/datagouv-components/package.json
@@ -17,6 +17,10 @@
     ]
   },
   "main": "./src/main.ts",
+  "exports": {
+    ".": "./src/main.ts",
+    "./chart": "./src/chart.ts"
+  },
   "sideEffects": false,
   "files": [
     "assets",
--- a/datagouv-components/package.json
+++ b/datagouv-components/package.json
@@ -20,7 +20,8 @@
   "exports": {
     ".": "./src/main.ts",
     "./chart": "./src/chart.ts",
-    "./assets/*": "./assets/*"
+    "./assets/*": "./assets/*",
+    "./dist/*": "./dist/*"
   },
   "sideEffects": false,
   "files": [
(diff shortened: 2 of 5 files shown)


ThibaudDauce:
Why `echarts` is not installed? If it's used inside `@datagouv/components-next` it should be installed as a dependency for `@datagouv/components-next`, no?

abulte:
> Why `echarts` is not installed? If it's used inside `@datagouv/components-next` it should be installed as a dependency for `@datagouv/components-next`, no?

Good question! I'm not sure I get it completely, but looks like a combination of `source` install here + `pnpm` install downstream. I'm actually happy if it's not installed downstream for now.

BTW, it could be cleaner/more deterministic to declare echarts as a `peerDependencies` here. The charts endpoint would still make sense.

ThibaudDauce:
> > Why `echarts` is not installed? If it's used inside `@datagouv/components-next` it should be installed as a dependency for `@datagouv/components-next`, no?
> 
> Good question! I'm not sure I get it completely, but looks like a combination of `source` install here + `pnpm` install downstream. I'm actually happy if it's not installed downstream for now.
> 
> BTW, it could be cleaner/more deterministic to declare echarts as a `peerDependencies` here. The charts endpoint would still make sense.

I let @nicolaskempf57 decide, but I'm not a big fan of adding an entrypoint… It adds some asymmetries to the package for a random feature (charts).

abulte:
> > > Why `echarts` is not installed? If it's used inside `@datagouv/components-next` it should be installed as a dependency for `@datagouv/components-next`, no?
> > 
> > 
> > Good question! I'm not sure I get it completely, but looks like a combination of `source` install here + `pnpm` install downstream. I'm actually happy if it's not installed downstream for now.
> > BTW, it could be cleaner/more deterministic to declare echarts as a `peerDependencies` here. The charts endpoint would still make sense.
> 
> I let @nicolaskempf57 decide, but I'm not a big fan of adding an entrypoint… It adds some asymmetries to the package for a random feature (charts).

Not so random, `echarts` is _huge_.

abulte:
@ThibaudDauce @nicolaskempf57 after digging a bit, the original issue is strictly a local one (yalc installed package does not install echarts). I still think it's useful to make the echarts dep optional (and it fixes the original bug), so I changed the scope of this PR.

nicolaskempf57 (approved):
I don't mind having a separate export for charts, seems to be a nice DX quick win

PR #1096, opened 2026-05-25
feat(search|a11y): announce results count, optional autofocus

abulte:
Two features to improve a11y downstream:
- optional autofocus on search input field (udata-front-kit focuses on a11y quick links)
- emit an event with results count (for screen readers announcements)

### DIFF
--- a/datagouv-components/src/components/Search/GlobalSearch.vue
+++ b/datagouv-components/src/components/Search/GlobalSearch.vue
@@ -13,7 +13,7 @@
       <SearchInput
         v-model="q"
         :placeholder="resolvedPlaceholder"
-        :auto-focus="props.autoFocus"
+        :auto-focus
       />
     </div>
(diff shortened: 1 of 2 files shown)


datagouv-components/src/components/Search/GlobalSearch.vue
```diff
@@ -13,6 +13,7 @@
       <SearchInput
         v-model="q"
         :placeholder="resolvedPlaceholder"
+        :auto-focus="props.autoFocus"
```
ThibaudDauce:
```suggestion
        :auto-focus
```

ThibaudDauce (approved):

abulte:
@ThibaudDauce I don't think we need #1095 for this to work with a production build (I just can't test locally), so I rebased on main and I'm keen on testing the dev version on udata-front-kit when it's merged.

PR #1098, opened 2026-05-27
fix: sending credentials with new CORS policy

ThibaudDauce:
### Problem

The backend (`udata`) hardened its CORS handling: only trusted origins (the cdata front) now receive `Access-Control-Allow-Origin: <origin>` + `Access-Control-Allow-Credentials: true`. Any other origin — e.g. the `udata-front-kit` verticals (`simplifions`, `écosphères`…) — now gets `Access-Control-Allow-Origin: *` without credentials.

`datagouv-components`' default fetch sent every request with `credentials: 'include'`, and the browser forbids `Access-Control-Allow-Origin: *` together with `credentials: 'include'`. As a result, every component request from the verticals (e.g. `GlobalSearch` hitting `/api/2/topics/search/`) failed with:

> The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.

The verticals authenticate via an `Authorization: Bearer` header, not the cookie, so `credentials: 'include'` was both useless and harmful for them. cdata itself was unaffected (it provides a `customUseFetch` and is a trusted origin).

### Changes (`datagouv-components`)

- **Removed the hardcoded `credentials: 'include'`** from the default `useFetch` and from the metrics CSV export (`createDatasetsForOrganizationMetricsUrl`), which were the two shared code paths breaking CORS cross-origin.
- **Introduced a single configured `$fetch`** in the plugin config, the single source of truth for requests (auth, headers, error handling):
  - The plugin `install` defaults it to an `ofetch` instance built from the existing `onRequest`/`onResponse` config hooks (baseURL, JSON headers, `X-API-KEY`, `lang`, plus the consumer's hooks). So `config.$fetch` is **always defined**, and `useComponentsConfig()` types it as non-optional — no `?? ofetch` fallback needed at any call site.
  - The default `useFetch` now just delegates to `config.$fetch` (the per-call `onRequest` layer is gone from `api.ts`). `raw` requests stay on bare `ofetch` since they target another data.gouv service (the Tabular API) with its own absolute URL.
  - Imperative helpers (metrics CSV export) call `config.$fetch` directly.

### Changes (cdata)

- `app.vue` / `error.vue` now pass `$fetch: $api` alongside the existing `customUseFetch: useAPI`, so the metrics CSV export goes through cdata's authenticated `$api` (cookie + token + error handling) instead of a raw cookie-based `ofetch`.

### Backward compatibility

Non-breaking. Verticals that still configure `onRequest` (their Bearer injection) keep working unchanged: the `install` folds those hooks into the default `$fetch`. They can later switch to providing their own `$fetch` and drop the hooks, but it's not required.

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -21,6 +21,7 @@
 import 'vue-sonner/style.css'
 import { Toaster, datagouv } from '@datagouv/components-next'
 import type { UseFetchFunction } from '@datagouv/components-next'
+import type { $Fetch } from 'ofetch'
 import CdataLink from './components/CdataLink.vue'
 import { ClientOnly, TextClamp } from '#components'
 
@@ -52,6 +53,7 @@ app.vueApp.use(datagouv, {
   tabularApiDataserviceId: runtimeConfig.public.tabularApiDataserviceId,
   tabularAllowRemote: true,
   customUseFetch: useAPI as UseFetchFunction, // Why this `as` is required?
+  $fetch: app.$api as $Fetch, // Imperative fetch for the lib's non-reactive helpers (metrics CSV export). Cast: Nuxt's $Fetch type lacks `native`, like the `as UseFetchFunction` above.
   textClamp: TextClamp,
   appLink: CdataLink,
   clientOnly: ClientOnly,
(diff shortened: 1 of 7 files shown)


abulte (approved):
Works fine locally with a vertical. I didn't check the code thoroughly.

PR #1099, opened 2026-05-28
feat: add nuxt module for robots.txt

ThibaudDauce:
From the documentation https://nuxtseo.com/docs/robots/guides/disable-indexing#handling-staging-environments we should add a `NUXT_SITE_ENV=dev` and `NUXT_SITE_ENV=demo`. The default is `production` and should allow indexation, as soon as we have a different value, indexing is disabled.

Draft: because I need to include back the existing `robots.txt` from `infra`.


maudetes:
I don't think we need to import the robots.txt in cdata for now, let's keep it into `infra`.

PR #1102, opened 2026-05-29
fix: use reuses x-fields in reuses tab

nicolaskempf57:
This PR prevent the reuses tab on the dataset page to fetch the dataset data every time and all its ressources


maudetes (approved):
Thank you!
Not sure it will be needed once https://github.com/datagouv/data.gouv.fr/issues/2007 is done

PR #1106, opened 2026-06-01
feat(search): restrict org filter to orgs with an API on dataservices…

Samuelfaure:
… search

Pass dataservices=true to the org suggest endpoint when the active search is the dataservices (API) search, so the organization filter only offers orgs that have at least one API. Implements API2-314 (frontend half).

fix https://linear.app/pole-api/issue/API2-314/retirer-les-organisations-par-defaut-du-filtre-organisation-lorsque

linked with https://github.com/opendatateam/udata/pull/3804

### DIFF
--- a/datagouv-components/src/components/Form/OrganizationSelect.vue
+++ b/datagouv-components/src/components/Form/OrganizationSelect.vue
@@ -36,8 +36,9 @@ import SearchableSelect from './SearchableSelect.vue'
 const model = defineModel<Organization | OrganizationSuggest | null>({ default: null })
 const id = defineModel<string | undefined>('id')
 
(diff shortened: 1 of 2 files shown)


maudetes:
There are also similar discussions for using organization facets [here](https://github.com/datagouv/data.gouv.fr/issues/1432)

Samuelfaure:
@maudetes Interessant, je vais faire une repasse

Samuelfaure:
@maudetes After looking into it, I propose we ship this as-it and switch to facets in a later PR

It's too big of a change for the scope of this PR so let's tackle it in a subsequent bigger PR

PR #1108, opened 2026-06-01
feat: clarify tabular API access

Pierlou:


components/Datasets/FileEditModal.vue
```diff
@@ -150,7 +150,7 @@
                 <p>
                   {{ $t('Attention : cette ressource est exposée via une') }}
                   <a
-                    href="https://www.data.gouv.fr/dataservices/api-tabulaire-data-gouv-fr-beta"
+                    href="https://www.data.gouv.fr/dataservices/673b0e6774a23d9eac2af8ce"
```
nicolaskempf57: We have the id in `runtimeConfig.public.tabularApiDataserviceId`, also we might want to use `siteConfig.url` instead of `https://www.data.gouv.fr`

components/Datasets/FileEditModal.vue
```diff
@@ -150,7 +150,7 @@
                 <p>
                   {{ $t('Attention : cette ressource est exposée via une') }}
                   <a
-                    href="https://www.data.gouv.fr/dataservices/api-tabulaire-data-gouv-fr-beta"
+                    :href="`${siteConfig.url}/dataservices/${runtimeConfig.public.tabularApiDataserviceId}`"
                     target="_blank"
                     class="fr-link"
                   >
```
ThibaudDauce: Could be a simple relative NuxtLink , no?

ThibaudDauce (approved):

PR #1109, opened 2026-06-01
feat(search): customizable type icons

abulte:
Lets downstream pass an icon to the types list of GlobalSearch. The icon can be a string (we're not using RemixIcons in the front-kit, so we just pass a dsfr icon class name).


ThibaudDauce (approved):
icon as string is so 2000 but why not :-D

PR #1110, opened 2026-06-02
feat: add topic page

ThibaudDauce:
- Require https://github.com/opendatateam/udata/pull/3765

image

### DIFF
--- a/pages/topics/[id]/index.vue
+++ b/pages/topics/[id]/index.vue
@@ -80,6 +80,12 @@ const reusesQuery = computed(() => ({
   page_size: reusesPageSize.value,
   topic: props.topic.id,
 }))
+// We use the search endpoint for reuses because there is no v2 reuses index
+// filtered by topic yet. Once udata#3800 is merged we can switch to the faster
+// v2 index, like we already do for datasets above.
+// If we ever need the TopicElement data (title/description/extras attached to
+// the element), we could instead query /api/2/topics/{id}/elements/ for both
+// datasets and reuses.
 const { data: reuses } = await useAPI<PaginatedArray<Reuse>>('/api/2/reuses/search/', {
   headers: { 'X-Fields': reusesXFields },
   query: reusesQuery,
--- a/pages/topics/[id]/index.vue
+++ b/pages/topics/[id]/index.vue
@@ -1,0 +1,87 @@
+<template>
+  <div class="space-y-8">
+    <MarkdownViewer
+      v-if="topic.description"
+      class="w-full min-w-0"
+      :content="topic.description"
+      :min-heading="3"
+    />
+
+    <section v-if="datasets && datasets.total">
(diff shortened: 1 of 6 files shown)


pages/topics/[id]/index.vue
```diff
@@ -0,0 +1,87 @@
...
+}>()
+
+const datasetsPage = ref(1)
+const datasetsPageSize = ref(10)
+const datasetsQuery = computed(() => ({
+  page: datasetsPage.value,
+  page_size: datasetsPageSize.value,
+  topic: props.topic.id,
+}))
+const { data: datasets } = await useAPI<PaginatedArray<DatasetV2>>('/api/2/datasets/', { query: datasetsQuery })
```
Comment (maudetes) [inline pages/topics/[id]/index.vue:74]: datasets use `api/2/` list endopint when reuses use `search` endpoint? Shouldn't we use the same logic between the two?
We could also use `api/2/topics/<topic_id>/elements/?class=<Reuse,Dataset>`?

utils/discussions.ts
```diff
@@ -31,6 +34,9 @@ export function getSubjectPage(subject: DiscussionSubjectTypes) {
   if (subject === null) {
     return ''
   }
+  if ('elements' in subject) {
+    return subject.page
+  }
```
maudetes, line 43: I'm unsure about this check. Isn't the following condition enough?

maudetes (commented):
Thank you ✨

maudetes (approved):

pages/topics/[id]/index.vue
Comment (ThibaudDauce) [reply pages/topics/[id]/index.vue:74]: I think TopicElement will give us more information yes. But do we want to use this specific endpoint if we only show the cards?

Will look into why search vs index

utils/discussions.ts
ThibaudDauce: Yes it's enough when the udata PR is merged. It's only required to test right now.

pages/topics/[id]/index.vue
Comment (ThibaudDauce) [reply pages/topics/[id]/index.vue:74]: I think we don't want to use index for reuse since it's slow… But we could maybe wait for https://github.com/opendatateam/udata/pull/3800 to switch from search to index v2 for reuse… Not sure we want to use `/elements` if we don't use the TopicElement info?

utils/discussions.ts
ThibaudDauce: Commented in 03eefd32

pages/topics/[id]/index.vue
Comment (maudetes) [reply pages/topics/[id]/index.vue:74]: Yes, it's a bit early for showing elements info so we can keep this strategy, with reuse search for now.
Maybe add a comment so that we update later on when we have reuse v2 or we finally use elements?

Comment (ThibaudDauce) [reply pages/topics/[id]/index.vue:74]: Done in 15216c64

PR #1113, opened 2026-06-02
fix: hide stats download without datasets

### LINKED ISSUE #958: When user has no dataset, don't display button to download statistics
When user has no published dataset, don't display button to download statistics.

(currently when downloading monthly stats, it downloads an empty CSV file; and when downloading teh catalog it shows a 404)

Image

bolinocroustibat:
Fixes #958

When an organization has no datasets, download buttons on the metrics and datasets admin pages are now hidden instead of offering empty CSV downloads or 404 responses.

### DIFF
--- a/components/Datasets/AdminDatasetsPage.vue
+++ b/components/Datasets/AdminDatasetsPage.vue
@@ -43,8 +43,8 @@
         />
         <div>
           <BrandedButton
-            v-if="organization"
-            :href="pageData.total ? `${config.public.apiBase}/api/1/organizations/${organization.id}/datasets.csv` : undefined"
+            v-if="organization && organization.metrics.datasets > 0"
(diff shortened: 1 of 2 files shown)


nicolaskempf57 (approved):

components/Datasets/AdminDatasetsPage.vue
```diff
@@ -56,22 +56,22 @@
     </div>
 
     <LoadingBlock
-      v-slot="{ data: pageData }"
+      v-slot="{ data }"
```
nicolaskempf57, line 59: I don't think this change is necessary ?

bolinocroustibat: No but there were an alert on the PR otherwise

PR #1114, opened 2026-06-02
feat(search): customizable defaultSort

abulte:
This lets the downstream component set a `defaultSort` for a given type.

I chose to hide the default Pertinence option when a `defaultSort` is set. This avoids having to handle a new `relevance` placeholder value for this sort order: it currently uses `undefined` and would clash with `sort = undefined ?? defaultSort`. YMMV, not a problem downstream for now.

### DIFF
--- a/datagouv-components/src/components/Search/GlobalSearch.vue
+++ b/datagouv-components/src/components/Search/GlobalSearch.vue
@@ -200,10 +200,13 @@
               <div class="fr-col">
                 <select
                   id="sort-search"
-                  v-model="sort"
+                  v-model="effectiveSort"
                   class="fr-select text-sm shadow-input-blue!"
                 >
-                  <option :value="undefined">
+                  <option
+                    v-if="!currentTypeConfig?.defaultSort"
+                    :value="undefined"
+                  >
                     {{ t('Pertinence') }}
                   </option>
                   <option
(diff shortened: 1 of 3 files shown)


abulte:
(Tests are failing all over the place with a 30min timeout this afternoon 🤷)

nicolaskempf57 (approved):
This is working great ! I don't how if we can easily do it in TS but if both `sortOptions` and `defaultSort` are provided, `defaultSort` should be restricted to the values in `sortOptions`. 

I found it by mistake because I tried `reuses` for the dataset search config, and it's not in our `defaultDatasetSortOptions`

abulte:
> This is working great ! I don't how if we can easily do it in TS but if both `sortOptions` and `defaultSort` are provided, `defaultSort` should be restricted to the values in `sortOptions`.
> 
> I found it by mistake because I tried `reuses` for the dataset search config, and it's not in our `defaultDatasetSortOptions`

It should already be constrained by the type definitions and your IDE or type checker should complain, no?

```
sortOptions?: SortOption<DatasetSearchSort>[]
defaultSort?: DatasetSearchSort
```

nicolaskempf57:
Yes but it's only contrained by DatasetSearchSort. For exemple, in cdata we have sortOptions defined as 

```
  { value: '-created', label: 'Date de création' },
  { value: '-last_update', label: 'Dernière mise à jour' },
  { value: '-followers', label: `Nombre d'abonnés` },
  { value: '-reuses', label: 'Nombre de réutilisations' },
```

but DatasetSearchSort has other available values like `reuses` or `views`. If `defaultSort` is `reuse`, this is valid for the typechecker / IDE, but the corresponding UI is not working properly

abulte:
> but DatasetSearchSort has other available values like `reuses` or `views`. If `defaultSort` is `reuse`, this is valid for the typechecker / IDE, but the corresponding UI is not working properly

Right. I added a runtime warning on dev in https://github.com/datagouv/cdata/pull/1114/commits/f4bea4c6f11d4382632d6fe2726effa12cb2d989

Here is an LLM-assisted solution re types. I find it too complicated to reason about and I didn't find any easier solution, hence the console warning instead: 

```typescript
export function getDefaultDatasetConfig<S extends DatasetSearchSort = DatasetSearchSort>(
  overrides?: Partial<Omit<DatasetSearchConfig, 'class' | 'sortOptions' | 'defaultSort'>> & {
 sortOptions?: SortOption<S>[], defaultSort?: NoInfer<S> },
): DatasetSearchConfig
```

PR #1118, opened 2026-06-04
Feat: allow markdown on rate limiting field

Samuelfaure:
fix https://linear.app/pole-api/issue/API-6848/formater-correctement-les-liens-dans-les-champs-limite-dappels-et-taux

### DIFF
--- a/pages/dataservices/[did].vue
+++ b/pages/dataservices/[did].vue
@@ -150,10 +150,10 @@
                       {{ $t(`Limite d'appels`) }}
                     </dt>
                     <dd class="p-0 m-0">
-                      <span
+                      <MarkdownViewer
                         v-if="dataservice.rate_limiting"
-                        class="mr-1"
-                      >{{ dataservice.rate_limiting }}</span>
+                        :content="dataservice.rate_limiting"
+                      />
                       <a
                         v-if="dataservice.rate_limiting_url"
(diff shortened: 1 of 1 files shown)


nicolaskempf57:
We may also want to update the udata side to add `markdown=True` to the field

nicolaskempf57 (approved):

maudetes:
I don't have access to the ticket but if the issue is with the rendering on [this API for example](https://www.data.gouv.fr/dataservices/api-entreprise), we added a dedicated `rate_limiting_url` field in https://github.com/opendatateam/udata/pull/3262 and https://github.com/datagouv/cdata/pull/1005.
I don't think we want to add raw markdown on a rate limiting field?

Samuelfaure:
@maudetes I should have mentioned this but we do, discussed this with @DorineLam , we intend to integrate rate limiting as clickable URLs on dataservice pages

maudetes:
> @maudetes I should have mentioned this but we do, discussed this with @DorineLam , we intend to integrate rate limiting as clickable URLs on dataservice pages

See [an example](https://demo.data.gouv.fr/dataservices/api-entreprise) of what we can do when setting `rate_limiting_url`. Doesn't this answer the need?

Samuelfaure:
@DorineLam likes it, closing here 👍

PR #1121, opened 2026-06-04
feat(notifications): add localMarkAsRead and multi-page refresh

### LINKED ISSUE #2029: Notifications : Mettre à jour la liste des notifs coté front lors de l'action "Marquer comme lues" pour éviter de devoir refresh les notifications
Actuellement, marquer une notification comme lue provoque 2 choses :
- un appel d'API pour marquer la notification comme lue
- un refresh de la liste des notifications, qui recharge la première page

Cela pose problème dans le cas d'un utilisateur qui clique sur une notification en page 3, et qui "perd" ensuite le chargement des notifications de la page 2 et 3.

Une solution simple est de refaire le traitement coté front (ajouter la date comme handle_at), au lieu du refresh pour éviter des appels API

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/2029

This PR adds a new `localMarkAsRead`, that update the front-end notification data instead of refreshing it for every notification read. This is used when reading all notification without action, even with the `marquer comme lues` button.

It also update the refresh function to handle refreshing multiple pages. This is used after handling notifications with actions (when we sometime don't have a direct access to the notification object).

We could apply the `localMarkAsRead` everywhere but that imply doing some matching to find the notification linked to the current event, which seems more error-prone than the refresh.


composables/useMarkAsRead.ts
```diff
@@ -13,8 +13,9 @@ export function useMarkAsRead() {
 
     try {
       loading.value = true
+      localMarkAsRead(notification)
```
ThibaudDauce: If the API call fail we have an empty list but no information about the failure? Maybe not an issue…

composables/useNotifications.client.ts
```diff
@@ -33,19 +33,32 @@ export function useNotifications() {
   async function refreshNotifications() {
     start()
     try {
-      page.value = 1
-      const notifications = await $api<PaginatedArray<UserNotification>>('/api/1/notifications/', {
+      const toPage = page.value
+      notificationsCombinedList.value = []
```
ThibaudDauce: Will this flash the list for the user? Empty it then fill it again with an await below?

composables/useNotifications.client.ts
```diff
@@ -33,19 +33,32 @@ export function useNotifications() {
...
+      }
+      pendingNotifications.value = await $api<PaginatedArray<UserNotification>>('/api/1/notifications/', {
         params: {
-          page_size: PAGE_SIZE,
-          page: page.value,
+          page_size: 1,
+          page: 1,
+          handled: false,
         },
       })
```
ThibaudDauce: Do we want to merge this bloc with `refreshPendingNotifications` below?

components/Charts/ChartConfigurator.vue
```diff
@@ -293,6 +293,9 @@
                 <option value="avg">
                   {{ $t('Moyenne') }}
                 </option>
+                <option value="count">
+                  {{ $t('Compter') }}
+                </option>
               </select>
```
ThibaudDauce: To remove?

composables/useNotifications.client.ts
```diff
@@ -33,19 +33,32 @@ export function useNotifications() {
   async function refreshNotifications() {
     start()
```
ThibaudDauce: `loadNotifications()` already do a `start()` and `finish()` inside this `start()` is it working?

composables/useNotifications.client.ts
nicolaskempf57: With a slow network, I guess a user could open the notification list before the loading is done. But the list is not opened when this happens so for most users is it not an issue

composables/useNotifications.client.ts
nicolaskempf57: I'll try to remove the start/finish from refreshNotifications

composables/useMarkAsRead.ts
nicolaskempf57: I added 2 catches for refreshPendingNotifications and loadNotifications

composables/useNotifications.client.ts
```diff
@@ -25,31 +28,38 @@ export function useNotifications() {
...
   }
 
   async function refreshNotifications() {
+    const toPage = page.value
+    notificationsCombinedList.value = []
+    for (let p = 1; p <= toPage; p++) {
+      page.value = p
+      await loadNotifications()
+    }
+    refreshPendingNotifications()
```
ThibaudDauce:
```suggestion
    await refreshPendingNotifications()
```

composables/useNotifications.client.ts
```diff
@@ -1,5 +1,7 @@
 import type { UserNotification } from '~/types/notifications'
 import type { PaginatedArray } from '~/types/types'
+import { toast } from '@datagouv/components-next'
+import { useTranslation } from '~/composables/useTranslation'
```
ThibaudDauce: Is it useful? Not auto-imported?

composables/useMarkAsRead.ts
```diff
@@ -1,9 +1,9 @@
...
-import { canMarkAsRead } from '~/utils/notifications'
+import { canMarkAsRead, localMarkAsRead } from '~/utils/notifications'
 
 export function useMarkAsRead() {
   const loading = ref(false)
-  const { refreshNotifications } = useNotifications()
+  const { refreshPendingNotifications } = useNotifications()
   const { $api } = useNuxtApp()
 
   const markAsRead = async (notification: UserNotification) => {
```
ThibaudDauce, line 9: `markAsRead` do `localMarkAsRead` before the API call, `markWithoutActionAsRead` do the local after the API call. Maybe always do after the API call?

ThibaudDauce (approved):

PR #1125, opened 2026-06-08
fix: cors in dev to hit demo server

ThibaudDauce:


maudetes:
See [this existing PR](https://github.com/datagouv/cdata/pull/1122) as well

ThibaudDauce:
I think we don't need it in production so the if/else on devApiKey is useless, what do you think?

maudetes:
Yep agreed, but I'm unsure if other cdata may need it?
Or even if we want some day to have a different api domain?

ThibaudDauce:
> Yep agreed, but I'm unsure if other cdata may need it? Or even if we want some day to have a different api domain?

Yes me too, but we could add it back if needed? Maybe on a different config than just `devApiKey` anyway?

ThibaudDauce:
We need the condition for the CI and the local dev.local:3000 vs dev.local:7000 is not same-origin so we need to add the include: "credentials"

PR #1129, opened 2026-06-09
fix(perfs): remove duplicate API calls in new_explorer=1

ThibaudDauce:
The legacy accordion and the new duplicates the calls  to `https://demo.data.gouv.fr/api/2/datasets/5eebbc067a14b6fecc9c9976/resources/?type=documentation&page_size=10&lang=fr` for all types. This calls are slow so we don't want to do them twice (once is really enough).

Also fix the explore button to use the same URL than the accordion to allow Nuxt to de-duplicate the request too between components.

### DIFF
--- a/pages/datasets/[did].vue
+++ b/pages/datasets/[did].vue
@@ -606,10 +606,15 @@ onMounted(async () => {
   ])
 })
 
-const { data: resources } = await useAPI<PaginatedArray<Resource>>(
-  `/api/2/datasets/${route.params.did}/resources/`,
-  { query: { type: 'main' } },
-)
+// Use the same cache key as ResourceExplorer's `main` fetch (dataset id + identical params)
+// so Nuxt dedupes the two into a single request on the resources tab. `q: undefined` mirrors
+// the explorer's params exactly: without it the keys differ and both requests fire.
+const { data: resources } = dataset.value
+  ? await useAPI<PaginatedArray<Resource>>(
+      `/api/2/datasets/${dataset.value.id}/resources/`,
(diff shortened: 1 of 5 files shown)


pages/datasets/[did].vue
```diff
@@ -606,10 +606,19 @@ onMounted(async () => {
...
-const { data: resources } = await useAPI<PaginatedArray<Resource>>(
-  `/api/2/datasets/${route.params.did}/resources/`,
-  { query: { type: 'main' } },
-)
+// Use the same cache key as ResourceExplorer's `main` fetch (dataset id + identical params)
+// so Nuxt dedupes the two into a single request on the resources tab. Every part of the key
+// must stay byte-for-byte identical to the explorer's `mainParams`, otherwise the keys diverge
+// and both requests fire again (silent perf regression, no error, no failing test):
+//   - dataset id (not route.params.did, which can be a slug)
+//   - page_size: 10 must match ResourceExplorer's PAGE_SIZE constant (not exported, keep in sync)
```
Comment (nicolaskempf57) [inline pages/datasets/[did].vue]: could we use the same constant and export it somewhere ?

Comment (ThibaudDauce) [reply pages/datasets/[did].vue]: Yes but it creates a weird relation between the two components, and I wasn't sure it's wanted…

Comment (nicolaskempf57) [reply pages/datasets/[did].vue]: We can put it in `utils/datasets.ts` instead of one of the components

Comment (ThibaudDauce) [reply pages/datasets/[did].vue]: Done in 467f0b31

nicolaskempf57 (approved):

PR #1138, opened 2026-06-11
fix(resource): fix api tab for resource accordion

nicolaskempf57:
The ResourceExplorer is already working


Pierlou (approved):
Thanks for the fix 🙏

ThibaudDauce (approved):

PR #1149, opened 2026-06-17
fix(search): regression on type navigation

abulte:
#1148 broke the navigation between types with type-scoped filters, cf https://github.com/opendatateam/udata-front-kit/actions/runs/27694406152/job/81913620279 failing tests. This is an attempt to fix the behaviour but handling those filters differently.


abulte:
Handled downstream in https://github.com/opendatateam/udata-front-kit/pull/1265

PR #1162, opened 2026-07-02
feat(charts): improve count UX

### LINKED ISSUE #2023: Datastudio : gérer le cas "compter" pour la colonne Y

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/2023


components/Charts/ChartConfigurator.vue
```diff
@@ -579,24 +583,40 @@ const xAxisColumnOptions = computed<Array<XAxisColumnOption>>(() =>
...
-  return [
+  const sameColumnAsX = xAxisColumn && yAxisColumn === xAxisColumn
+  const options: SortOption[] = [
     { value: '', column: '', direction: '', label: t('Aucun'), icon: null },
     { value: 'axis_x-asc', column: xAxisColumn || t('Axe X'), direction: 'asc', label: t('Ascendant'), icon: RiArrowUpLine },
     { value: 'axis_x-desc', column: xAxisColumn || t('Axe X'), direction: 'desc', label: t('Descendant'), icon: RiArrowDownLine },
-    { value: 'axis_y-asc', column: yAxisColumn || t('Axe Y'), direction: 'asc', label: t('Ascendant'), icon: RiArrowUpLine },
-    { value: 'axis_y-desc', column: yAxisColumn || t('Axe Y'), direction: 'desc', label: t('Descendant'), icon: RiArrowDownLine },
   ]
+  if (!sameColumnAsX) {
```
ThibaudDauce, line 616: When Y == X (count mode) the `axis_y-*` options are dropped from `sortOptions`, but a previously selected `axis_y-asc`/`axis_y-desc` is never cleared from `form.x_axis.sort_combined` (the `sortProxy` setter only writes on a non-null value). `sortProxy.get` then returns null so the UI shows the placeholder, while the stale value survives and `toChartApi` still persists it as `sort_x_by: 'axis_y'` on a chart that has no distinct Y axis. Reset `sort_combined` to `''` when its option disappears (e.g. when switching to count).

ThibaudDauce (approved):

PR #1163, opened 2026-07-03
feat: compute serie type and min / max

### LINKED ISSUE #2025: Datastudio : calculer nous même le type (continu ou discret)
On essaye de changer la valeur par défaut à utiliser selon la détection, mais on garde le champ éditable par les utilisateurices.

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/2025


datagouv-components/src/components/Chart/ChartViewer.vue
```diff
@@ -168,6 +178,8 @@ const echartsOption = computed(() => {
     xAxis: {
       type: mapXAxisType(props.chart.x_axis),
       name: (props.chart.x_axis as XAxis).column_x,
+      min: xAxisColumn.value?.min ?? undefined,
```
ThibaudDauce: `min`/`max` are set unconditionally, but the axis becomes `category` whenever the user switches a numeric column to `discrete`. On a category axis ECharts reads `min`/`max` as category indices, not value bounds, which breaks the axis. Only set them when `mapXAxisType(props.chart.x_axis) === 'value'`.

datagouv-components/src/functions/charts.ts
```diff
@@ -74,6 +74,18 @@ export function buildColumnsFromProfile(profile: { profile: TabularProfile }): A
...
+      type: colType,
+      min: normalizeProfileNumber(colProfile?.min),
+      max: normalizeProfileNumber(colProfile?.max),
+    }
   })
 }
+
+function normalizeProfileNumber(value: number | string | undefined): number | undefined {
+  if (value === undefined || value === null) return undefined
+  const parsed = Number(value)
```
ThibaudDauce, line 93: For `date` columns the profile `min`/`max` are ISO strings, so `Number(...)` returns `NaN` and the bounds are dropped. Since a date column defaults to `continuous`, the min/max feature never applies to date axes. Convert dates to timestamps if bounds should apply there.

ThibaudDauce (approved):

PR #1170, opened 2026-07-17
feat(cada): add minimal cada explore

maudetes:
The idea is to replace the existing https://cada.data.gouv.fr/ with a minimal exploration in cdata `/explore/cada`.

Replace a dedicated Flask app x mongoDB x ES with a TabularExplorer component (based on [cada Tabular API](https://tabular-api.data.gouv.fr/api/resources/3a6d6fe3-8548-43ed-ad09-72052771447c/)) and detailed pages.

The homepage shows as an explore view. The generic popover is deactivated. The row `Numéro de dossier` has a link to the detailed Avis page.
image

The Avis page in `/explore/cdata/[id]` mimics previous behavior in cada.data.gouv.fr.
image

image

Some functionality loss:
* no facets
* no full text search, we use `__contains` for text (which means that there is no text normalization) and `__exact` for numbers (Numéro de dossier)
* multiple fields are comma separated, badges are split on comma on a detailed Avis page, but explore behaviour consider these as full string
* themes or conclusions are also hierarchical (on top of comma-separated), we show these as raw `Agriculture/Secteurs+Économiques` as they appear in the data

For easier redirect, we keep the filter param names, ie `meaning=Favorable/Sauf vie privée`, `administration=Mairie+de+Paris`, etc.

Test it locally with
```
NUXT_PUBLIC_CADA_RESOURCE_ID="3a6d6fe3-8548-43ed-ad09-72052771447c"
```
Or go to https://demo.data.gouv.fr/explore/cada


ThibaudDauce:
Why disable popover?

maudetes:
Thank you for the cleaning and update!

Regarding `json` type, I think a column with `json` should not use `contains` either. See column filter on `coordonneesXY` in http://dev.local:3000/explore?resource_id=0d562af2-d8c1-4e91-b1bf-c28f54e31416

ThibaudDauce:
> Thank you for the cleaning and update!
> 
> Regarding `json` type, I think a column with `json` should not use `contains` either. See column filter on `coordonneesXY` in http://dev.local:3000/explore?resource_id=0d562af2-d8c1-4e91-b1bf-c28f54e31416

I have a new batch of updates in progress :-) Concerning the JSON I will look into this resource but doesn't seems to be linked to CADA? And Claude seems to tell that there was some regressions with the JSON changes, I may do another PR on the subject.

maudetes:
> Concerning the JSON I will look into this resource but doesn't seems to be linked to CADA?

Yes, but it was the same issue as filtering on a `Nombre` column (`Année` or `Numéro de dossier` for CADA).
It can totally be fixed in a dedicated PR, taking a look at all existing types.

maudetes (commented):
@ThibaudDauce it's good for me if you want to approve 🎉

ThibaudDauce (approved):

PR #1176, opened 2026-07-30
feat(sitemap): use udata generated sitemap

maudetes:
Following https://github.com/opendatateam/udata/pull/3830

Both udata and cdata sitemaps index will be exposed in robots.txt.


tests/sitemap.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  test('includes the external sitemap URL when NUXT_SITEMAP_INDEX_URL is set', async ({ page }) => {
+    test.skip(!SITEMAP_URL, 'NUXT_SITEMAP_INDEX_URL not set')
+
+    const text = await (await page.request.get('/sitemap_index.xml')).text()
+    expect(text).toContain(`<loc>${SITEMAP_URL}</loc>`)
+    const locCount = (text.match(/<loc>/g) || []).length
+    expect(locCount).toBe(3)
+  })
+
+  test('omits the external sitemap URL when NUXT_SITEMAP_INDEX_URL is not set', async ({ page }) => {
```
maudetes: I don't think it's worth adding a step to test it explicitly in CI?

ThibaudDauce:
Claude is not sure it's allowed to reference an index inside an index, but it may be better to just reference two sitemaps in our robots.txt (in infra)

Something like:
```
Sitemap: {{ udata__cdata_base_url }}/sitemap_index.xml
Sitemap: {{ udata__sitemap_index_url }}
```

maudetes:
> Claude is not sure it's allowed to reference an index inside an index, but it may be better to just reference two sitemaps in our robots.txt (in infra)

Hmm, indeed when looking it up again, I find contradictory information on nested sitemaps :sweat: 

Let's go for the robots approach, I liked the single entry point but it's not worth the trouble.

I've removed any reference to udata sitemap in https://github.com/datagouv/cdata/pull/1176/changes/8113fe8e7a19712407371f41c8e1f6de496a4374.

ThibaudDauce (approved):

PR #1191, opened 2026-08-24
feat: cache /api/1/site instead of calling it on every render

ThibaudDauce:
[GET /api/1/site](https://www.data.gouv.fr/api/1/site) is the 3rd most called endpoint in udata with 47M calls. I think we can cache it a lot but I only put 1min to start.

### DIFF
--- a/pages/datasets/[did].vue
+++ b/pages/datasets/[did].vue
@@ -644,14 +644,14 @@ const exploreHref = computed(() => {
 // The badge labels are a reference list needed only to name the badges this
 // dataset carries: most datasets carry none, and fetching it for them would be
 // one request per page view for nothing.
-const badgeTranslations = dataset.value?.badges?.length
-  ? (await useFetch('/nuxt-api/dataset-badges')).data
-  : null
+const { data: badgeTranslations } = await useFetch('/nuxt-api/dataset-badges', {
+  immediate: Boolean(dataset.value?.badges?.length),
+})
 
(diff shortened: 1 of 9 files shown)


pages/datasets/[did].vue
```diff
@@ -671,9 +674,14 @@ const datasetDownloadsResources = computed(() => datasetMetrics.value?.downloads
 const datasetVisitsTotal = computed(() => datasetMetrics.value?.visitsTotal ?? 0)
 const datasetDownloadsResourcesTotal = computed(() => datasetMetrics.value?.downloadsTotal ?? 0)
 
-const { data: categories } = await useAPI<Array<{ value: string, label: string, definition: string }>>('/api/1/access_type/reason_categories/')
+// Same reasoning as the badge labels above: only a restricted dataset names a
+// reason category, which is 0.3% of them.
+const reasonCategories = dataset.value?.access_type_reason_category
```
Comment (nicolaskempf57) [inline pages/datasets/[did].vue]: Do we want to use the same immediate key as in the followers PR (https://github.com/datagouv/cdata/pull/1192) instead ?

Comment (ThibaudDauce) [reply pages/datasets/[did].vue]: Good catch! Done in [ecbbb07](https://github.com/datagouv/cdata/pull/1191/commits/ecbbb07c3e979aefee012134dcb2fe47a0dde396)

nicolaskempf57 (approved):

PR #1198, opened 2026-08-26
fix: render every date through a `FormattedDate` component so timezone and clock differences stop reporting hydration mismatches

ThibaudDauce:
Maybe fix some issues reported by Sentry (hydratation mismatch that crash with `emitsOptions` on `null`)

Add correct `datetime` HTML element to provide the absolute URL and title on hover.


datagouv-components/src/components/FormattedDate.vue
```diff
@@ -0,0 +1,63 @@
...
+
+// Hovering a date gives the exact moment, which the displayed form usually omits:
+// the relative forms drop it entirely, and most styles drop the time.
+const title = computed(() => parsed.value
+  ? formatDate(parsed.value, { dateStyle: 'long', timeStyle: 'short' })
+  : undefined)
+
+// `formatDate` fills in a default `dateStyle` on the object it receives, so it gets a
+// copy: mutating `props.options` would leak the default back to the parent.
+const formatted = computed(() => {
```
nicolaskempf57, line 66: Could formatted use parsed.value instead of props.date ? Or add a guard before the first if ? 

It would prevent passing non-date but truthy strings (like `"not a date")` and avoid a `RangeError: date value is not finite in DateTimeFormat.format()` when we try to format a `new Date(NaN)`

nicolaskempf57 (approved):

datagouv-components/src/components/FormattedDate.vue
ThibaudDauce: Good catch! Done in c294328d

PR #1199, opened 2026-08-26
fix: answer a 404 instead of a 500 on a listing page that does not exist

ThibaudDauce:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/299587

Error 
GET /organizations/region-hauts-de-france/datasets?page=127
Invalid array length


tests/organizations/pagination.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  // - without it (this e2e environment), the search falls back to mongo, whose
+  //   paginator does `abort(404)` as soon as a page is empty and is not page 1
+  //   (udata/flask_mongoengine/pagination.py). The listing then fails to load, the
+  //   guard has no total to compare against, and the page stays a 200 showing a
+  //   loading error.
+  //
+  // Re-enable once the e2e backend runs elasticsearch. Until then the crash itself is
+  // covered by the getVisiblePages unit tests.
+  //
+  // test('a page past the last one answers a clean 404', async ({ page }) => {
```
nicolaskempf57:
```suggestion
  // test.fixme('a page past the last one answers a clean 404', async ({ page }) => {
```

We can use fixme instead of commenting if we want playwright to skip the test

Also, we should probably fix the test or remove it if it's unnecessary no.

ThibaudDauce: I want to make it work sometime, either by fixing udata to return the same thing in both paths or by using ES in E2E tests… But not sure what's the correct way for the API to respond if you have an opinion.

Will use `test.fixme`

ThibaudDauce: Done in 74dd1391

nicolaskempf57: I think I prefer the `API answers 200 with the total, so the guard computes the page count and answers a 404.` case

nicolaskempf57 (approved):

PR #1210, opened 2026-08-27
feat(charts): show charts cards in admin

### LINKED ISSUE #2021: Datastudio : Afficher les cartes de viz dans l'admin

nicolaskempf57:
Closes https://github.com/datagouv/data.gouv.fr/issues/2021

## Show chart cards in the site administration

- New `/admin/site/charts` page (superadmin) listing all charts as cards with pagination, plus a "Graphiques" entry in the admin sidebar. Cards link to the edit page `/admin/beta/charts/{id}`.
- New `ChartCard` component in `@datagouv/components-next` (image or placeholder, title, producer, update date, views, description), with an `xl` media size on `ObjectCard`, a `Chart` type on `Placeholder` and an `image` field on the `Chart` type. Also showcased on `/design/cards`.
- Chart edit page refactored: the page loads the chart (with `LoadingBlock`) and passes it to `ChartConfigurator` via an `initialChart` prop instead of `chartId`.
- Admin fetches use `server: false` (`/admin/**` is server-rendered; SSR fetches bypass e2e mocks and caused hydration mismatches — same pattern as the metrics pages).
- Tests: new e2e specs for the admin page and the card, plus chart-configurator tests for loading an existing chart (waiting for the echarts canvas before saving to fix a flaky image upload). `tests/visualizations/` is 42/42 green on chromium, typecheck and lint pass.


pages/admin/beta/charts/[cid].vue
```diff
@@ -0,0 +1,46 @@
...
+    </Breadcrumb>
+    <LoadingBlock
+      v-slot="{ data: chart }"
+      :status
+      :data="loadedChart"
+    >
+      <ChartConfigurator
+        v-if="chartForm"
+        v-model="chartForm"
+        :initial-chart="chart"
```
Comment (ThibaudDauce) [inline pages/admin/beta/charts/[cid].vue:22]: It seems the component is not using the `v-model` and recomputing the form based on the `initial-chart`, no?

tests/visualizations/admin-site-charts.spec.ts
ThibaudDauce, line 1: Why mock the API here rather than create real charts? `setupChart` in
chart-configurator.spec.ts already builds one through the UI, and the existing specs
POST then delete for cleanup — `tests/posts/kind.spec.ts:98` does exactly this against
`/admin/site/posts`, which stays server-rendered.

If the reason is that the fixtures cover states the dev backend won't produce easily
(one chart with an image, one without, a views count), that's fair — but it's worth a
comment here, because it's what forces `server: false` on the page and that's a
production behaviour change nothing else explains.

datagouv-components/src/components/ChartCard.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <ObjectCard media-size="xl">
```
ThibaudDauce, line 2: Missing the `#badge` slot for `chart.private` — the other cards all have it
(ReuseHorizontalCard.vue:3-9). Sysadmins see private charts in this list with nothing
setting them apart.

components/Charts/AdminChartsPage.vue
```diff
@@ -0,0 +1,81 @@
...
+            :chart-url="`/admin/beta/charts/${chart.id}`"
+          />
+        </div>
+        <Pagination
+          :page="page"
+          :page-size="pageSize"
+          :total-results="chartsPage.total"
+          @change="(changedPage: number) => page = changedPage"
+        />
+      </div>
```
ThibaudDauce, line 42: Is there a reason to go for cards where all the other admin pages use tables?

components/Charts/ChartConfigurator.vue
```diff
@@ -845,12 +801,6 @@ async function loadChart(id: string) {
   }
```
ThibaudDauce: With the fetch moved to the page, can anything still reach this catch? The two loaders and
the `$chartsApi` block all swallow their own errors, so the toast looks unreachable now —
`LoadingBlock` handles the failure case instead.

tests/visualizations/admin-site-charts.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+  ...histogramChart,
+  id: 'chart-line',
+  title: 'Courbe de démonstration',
+  slug: 'courbe-de-demonstration',
+  description: 'Une courbe de démonstration pour la carte.',
+  page: 'http://dev.local:3000/visualizations/courbe-de-demonstration/',
+  image: null,
+  series: [{ type: 'line', column_y: 'valeur', aggregate_y: 'sum', resource_id: 'resource-1', column_x_name_override: null, filters: null }],
+  metrics: { views: 7 },
+}
```
ThibaudDauce: These lines are copy/paste? Could we extract a fixture or a helper?

components/AdminSidebar/AdminSidebarMenu/AdminSidebarMenu.vue
```diff
@@ -150,6 +150,11 @@
               :label="$t('Réutilisations')"
               to="/admin/site/reuses"
             />
+            <AdminSidebarLink
+              :icon="RiBarChartBoxLine"
```
ThibaudDauce: `RiBarChartBoxLine` is already used for statistics, maybe we should choose a new icon for one or another

ThibaudDauce (approved):

components/Charts/AdminChartsPage.vue
nicolaskempf57: I think we are using cards instead of a table to show the images

ThibaudDauce: Reuses have images too but are shown with a table, not mandatory but was just wondering if it's wanted to change the admin pattern

pages/admin/beta/charts/[cid].vue
Comment (nicolaskempf57) [reply pages/admin/beta/charts/[cid].vue:22]: True ! I'm updating so that initial-chart is only use for the first producer / resource / dataset

components/Charts/AdminChartsPage.vue
nicolaskempf57: True, I think this is because the chart image is a snapshot and the reuse one is just a user-provided one

tests/visualizations/admin-site-charts.spec.ts
nicolaskempf57: Now it's using the form when possible or API calls (ex: for private charts)

tests/visualizations/admin-site-charts.spec.ts
nicolaskempf57: Now using createChart

datagouv-components/src/components/ChartCard.vue
```diff
@@ -0,0 +1,102 @@
...
+        :owner="chart.owner"
+        :organization-url="organizationUrl"
+      />
+      <RiSubtractLine
+        aria-hidden="true"
+        class="size-4 flex-none fill-gray-medium"
+      />
+      <p class="text-sm whitespace-nowrap mb-0 text-gray-medium">
+        {{ t('Mis à jour {date}', { date: formatRelativeIfRecentDate(chart.last_modified, { dateStyle: 'medium' }) }) }}
+      </p>
```
ThibaudDauce: You can now use FormattedDate from https://github.com/datagouv/cdata/pull/1198

pnpm-lock.yaml
ThibaudDauce, line 1: Why this change?

ThibaudDauce (approved):

pnpm-lock.yaml
nicolaskempf57: I broke my local env and I had to do install / reinstall to have it back, it shouldn't be committed

nicolaskempf57: Changes discarded during conflict resolution

PR #1214, opened 2026-09-02
feat: add explore feedback query params

nicolaskempf57:


pages/datasets/[did]/index.vue
```diff
@@ -39,14 +40,32 @@
...
+// resource shown in the explorer (forwarded via `select`, unknown during SSR)
+// and a simplified "Browser - device" string (client-side only, so it appears
+// after hydration).
+const feedbackBaseUrl = useRuntimeConfig().public.explorerFeedbackUrl
+const feedbackResource = ref<Resource | null>(null)
+const feedbackUserAgent = ref<string | null>(null)
+onMounted(() => {
+  feedbackUserAgent.value = getSimplifiedUserAgent(navigator.userAgent)
+})
+const feedbackUrl = computed(() => {
```
Comment (maudetes) [inline pages/datasets/[did]/index.vue:60]: I think it currently doesn't show in the fields in the form. As discussed, I think it would be fair to be transparent about the context information being sent

utils/explorer-feedback.ts
```diff
@@ -0,0 +1,77 @@
...
+  else if (ua.includes('mobile')) {
+    device = 'mobile'
+  }
+  else if (ua.includes('android')) {
+    // Android tablets usually omit both "Mobile" and "Tablet".
+    device = 'tablet'
+  }
+
+  return `${browser} - ${device}`
+}
```
ThibaudDauce, line 26: Why not use a simple package like https://www.npmjs.com/package/bowser ? There is a few bugs in the detection and it adds also a few useless tests on parsing agents…

pages/datasets/[did]/index.vue
Comment (nicolaskempf57) [reply pages/datasets/[did]/index.vue:60]: @agarrone regarding the form fields visibility

PR #1215, opened 2026-09-03
feat(resource): add external url override to ResourceViewer

abulte:
Related https://github.com/opendatateam/udata-front-kit/pull/1373

Allows to override the resource external url used for "Copier le lien" target. Useful for front-kit verticals.


datagouv-components/src/config.ts
```diff
@@ -45,6 +47,8 @@ export type PluginConfig = {
   clientOnly?: Component | null
   searchDebounce?: number
   forumUrl?: string
+  /** Overrides the "Copier le lien" target built by `getResourceExternalUrl`. */
+  resourceExternalUrl?: (dataset: Dataset | DatasetV2 | Omit<Dataset, 'resources' | 'community_resources'>, resource: Resource | CommunityResource) => string
```
ThibaudDauce: Maybe getResourceExternalUrl since it's a callback? And you can maybe add the information in the comment about absolute vs relative. I think it must be absolute?

ThibaudDauce (approved):

PR #1232, opened 2026-09-14
feat(chart): add chart filter groups

nicolaskempf57:


components/Charts/ChartConfigurator.vue
```diff
@@ -828,46 +863,36 @@ async function saveChart() {
...
-    form.value.filter.filters[index] = newFilter
-  }
+function addCondition(groupIndex: number) {
+  const groups = filterGroups.value.groups.map((g, i) => (i === groupIndex ? [...g, newEmptyFilter()] : g))
+  applyFilterGroups({ ...filterGroups.value, groups })
 }
 
-function addFilter() {
-  const newFilter: Filter = { _cls: 'Filter', column: '', condition: 'exact' as const, value: '' }
+function updateFilter(groupIndex: number, filterIndex: number, newFilter: Filter) {
```
ThibaudDauce: `updateFilter` is never called: `ChartFilterRow` edits the filter in place
(`v-model="filter.column"` and the two others), which doesn't trigger `update:modelValue`,
only reassigning the whole model would. So `@update:model-value` in ChartFilterGroup, the
`update:filter` emit and this function are all dead.

In-place editing looks intentional (`chartFilters.ts:18-20` says the groups keep the
original references, and the deep watcher refreshes the preview). Should we drop the
`update:filter` chain then? Right now the row is wired as a controlled component but
mutates through the reference, and both are in the same file.

components/Charts/ChartConfigurator.vue
```diff
@@ -137,25 +137,52 @@
...
+                />
+              </template>
+            </div>
+            <div class="flex flex-wrap gap-2">
+              <BrandedButton
+                v-if="totalRules === 0"
+                size="sm"
+                color="tertiary"
+                :icon="RiAddLine"
+                @click="addFilter"
```
ThibaudDauce: `addFilter` only runs when `totalRules === 0`, and there all groups are empty so
`fromFilterGroups` dropped them: it ends up doing the same as `addGroup`, one group with
one empty rule. The `push` into the last group is never reached.

```suggestion
                @click="addGroup"
```

`addFilter` can go then?

components/Charts/ChartFilterRow.vue
```diff
@@ -46,17 +55,27 @@ import { computed } from 'vue'
...
 }>()
 
 const { t } = useTranslation()
 
+const connectorLabel = computed(() => {
+  if (props.connector === 'first') return t('Quand')
+  if (props.connector === 'and') return t('et')
+  if (props.connector === 'or') return t('ou')
+  return ''
+})
```
ThibaudDauce: Looks like this is left over from the version where only the second row had the select
and the next ones showed a plain `et` / `ou`? Now that every row but the first has a
synced select, `connectorLabel` can only return `t('Quand')`: the span shows when
`showCombinatorSelect` is false, and `showCombinatorSelect && index > 0` (ChartFilterGroup
line 9) is just `index > 0`, since `index > 0` already means `totalRules > 1`.

Could the row take `index` back like on main, with `v-if="index > 0"` on the Listbox and
`Quand` hardcoded? Both props and the computed go away.

tests-unit/datagouv-components/tabular-api.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  it('serializes is_null inside a group without value', () => {
+    const filters: OrFilters = { _cls: 'OrFilters', filters: [f('a', 'is_null', null), f('b', 'exact', '1')] }
+    expect(buildFilterQuery(filters)).toBe('or=(a__isnull,b__exact.1)')
+  })
+
+  it('escapes dots and parens in values inside groups', () => {
+    const filters: OrFilters = { _cls: 'OrFilters', filters: [f('a', 'exact', '1.5'), f('b', 'exact', 'x(y)')] }
+    expect(buildFilterQuery(filters)).toBe('or=(a__exact.1%2E5,b__exact.x%28y%29)')
+  })
```
ThibaudDauce, line 59: This one is green but pins a string the API rejects: the `%2E` is decoded back to a `.`
before parsing, so `or=(a__exact.1%2E5)` is a 400 (see the other comment). The test name
says dots are escaped, which is what made me trust it at first.

```suggestion
  it('escapes dots and parens in values inside groups', () => {
    const filters: OrFilters = { _cls: 'OrFilters', filters: [f('a', 'exact', '1.5'), f('b', 'exact', 'x(y)')] }
    expect(buildFilterQuery(filters)).toBe('or=(a__exact.1%252E5,b__exact.x%2528y%2529)')
  })
```

components/Charts/ChartFilterRow.vue
```diff
@@ -46,17 +55,27 @@ import { computed } from 'vue'
 const filter = defineModel<Filter>({ required: true })
 
 const props = defineProps<{
-  index: number
+  connector: 'first' | 'and' | 'or' | 'none'
+  combinator: 'and' | 'or'
```
ThibaudDauce: `FilterGroupCombinator`, no? It's imported in ChartFilterGroup and used in the
configurator, this file is the only one rewriting the union.

components/Charts/ChartFilterRow.vue
```diff
@@ -46,17 +55,27 @@ import { computed } from 'vue'
...
+  showCombinatorSelect: boolean
   columnOptions: Array<{ key: string, value: string, disabled: boolean }>
   conditionOptions: Array<FilterCondition>
 }>()
 
 defineEmits<{
-  remove: []
+  'remove': []
+  'update:combinator': [combinator: 'and' | 'or']
```
ThibaudDauce: `FilterGroupCombinator`, no? It's imported in ChartFilterGroup and used in the
configurator, this file is the only one rewriting the union.

components/Charts/ChartFilterGroup.vue
```diff
@@ -0,0 +1,63 @@
...
+import type { FilterGroupCombinator } from '~/utils/chartFilters'
+import ChartFilterRow from './ChartFilterRow.vue'
+
+defineProps<{
+  group: Array<Filter>
+  combinator: FilterGroupCombinator
+  showCombinatorSelect: boolean
+  canRemoveGroup: boolean
+  bordered: boolean
+  columnOptions: Array<{ key: string, value: string, disabled: boolean }>
```
ThibaudDauce: Fourth copy of this shape (ChartConfigurator 633 and 634, ChartFilterRow 61, and here).
Maybe a named type in `utils/chartFilters.ts`? This file already imports from it.

ThibaudDauce (changes requested):

PR #1237, opened 2026-09-15
fix(structured): detect encoding

nicolaskempf57:


ThibaudDauce (changes requested):
The 64 KB sample is cut mid-character: when a multibyte char straddles byte 65536, `decode()` without `stream` throws and we fall back to `windows-1252`. So a valid UTF-8 CSV bigger than 64 KB comes out entirely mojibaked (`GÃ©nÃ©rateurs`, not `�`, so the new e2e assertion doesn't see it).

Repro file attached (the schema fixture repeated 17 times, valid UTF-8, the `é` of row 17 sits right on byte 65536).

[lave-linge-utf8-64ko.csv](https://github.com/user-attachments/files/32238255/lave-linge-utf8-64ko.csv)

ThibaudDauce (approved):

PR #1247, opened 2026-09-22
feat: matomo site search tracking

nicolaskempf57:


pages/organizations/[oid]/search.vue
```diff
@@ -12,13 +13,24 @@
...
 const route = useRoute()
 const searchType = ref<SearchType>((route.query.type as SearchType) || 'datasets')
 
+const { $matomo } = useNuxtApp()
+
+function trackSearch(keyword: string, type: string, total: number) {
+  if (!import.meta.client) return
+  $matomo.trackSiteSearch(keyword, type, total)
+}
+
```
Comment (ThibaudDauce) [inline pages/organizations/[oid]/search.vue]: Why use an event from `GlobalSearch`, I would prefer to have this in `GlobalSearch.vue` directly. I think it's to avoid having Matomo inside the datagouv-components but we could add an optional `Matomo` object inside the datagouv-components and do the tracking there? It would prevent the duplicate listeners in organisation and datasets and also I think `@search` event reads a little bit like the parent of this component should do something about the search?

ThibaudDauce (approved):

